PR #6, opened 2021-08-13
Switch deployment to GitHub Actions

parfeon:
## Switch deployment to GitHub Actions

Switch product deployment to GitHub Actions.


kleewho:
What about travis? Shouldn't we turn it off? Have you turn it off?

parfeon:
@kleewho public branch doesn't know, how to use it for release :)
At this moment Travis used only for tests and release will be done using GitHub Actions.

PR #9, opened 2021-11-05
SDK-1560 Revoke Token

bartk:
feat: Add revoke token feature


src/integrationTest/kotlin/com/pubnub/api/integration/pam/RevokeTokenIntegrationTest.kt
```diff
@@ -0,0 +1,43 @@
+package com.pubnub.api.integration.pam
```
kleewho, line 1: I'd rather like to see an acceptance tests ;) ;)

src/main/kotlin/com/pubnub/api/PubNubError.kt
```diff
@@ -131,6 +131,11 @@ enum class PubNubError(private val code: Int, val message: String) {
         "User name is missing"
     ),
 
+    TOKEN_MISSING(
+        152,
```
kleewho, line 135: I know that codes are not consistent across SDKs, but could you please check this number is free on Java SDK as well?

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/v3/PNRevokeTokenResult.kt
```diff
@@ -0,0 +1,3 @@
+package com.pubnub.api.models.consumer.access_manager.v3
+
+data class PNRevokeTokenResult(val token: String)
```
kleewho: It would be good to include what we have to return to the user in case of successful call in the specification document so we're consistent across the SDKs

src/main/kotlin/com/pubnub/api/services/AccessManagerService.kt
```diff
@@ -24,4 +27,11 @@ internal interface AccessManagerService {
         @Body body: Any?,
         @QueryMap options: Map<String, String>
     ): Call<GrantTokenResponse>
+
+    @DELETE("/v3/pam/{subKey}/grant/{token}")
+    fun revokeToken(
+        @Path("subKey") subKey: String,
+        @Path("token") token: String,
+        @QueryMap queryParams: HashMap<String, String>
```
kleewho, line 35: Please use `Map`

src/main/kotlin/com/pubnub/api/PubNubError.kt
kleewho: I've used the first free number on Java which is `168`

src/main/kotlin/com/pubnub/api/endpoints/access/RevokeToken.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    override fun createResponse(input: Response<RevokeTokenResponse>): Unit = Unit
+
+    override fun operationType(): PNOperationType = PNOperationType.PNAccessManagerRevokeToken
+    override fun isAuthRequired(): Boolean = false
+
+    private fun repairEncoding(token: String): String {
+        return token
+            .split(" ")
```
budgetpreneur, line 40: To my knowledge, token never comes with space. Any reason splitting token with space?

kleewho: Yes. A test scenario

kleewho: https://github.com/pubnub/sdk-specifications/blob/62e194d895dfac0e970dc2f6d9f3f76815de5b92/features/access/revoke-token.feature#L43

kleewho:
@budgetpreneur you're right. Copy paste error

kleewho (dismissed):

kleewho:
@client-engineering-bot release

techwritermat (approved):

kleewho (approved):

PR #18, opened 2022-03-11
retrofit licence link change

techwritermat:
changed the link because the old one had the `parent` part which was breaking the link


kleewho (approved):
Thank you!

PR #20, opened 2022-04-07
Remove v from release version to fix docs

kazydek:
This PR is removing `v` from the release version and, thus, fixing the Kotlin SDK docs to which the release version is automatically propagated:
Screenshot 2022-04-07 at 14 06 59

(should be `7.0.0` instead of `v7.0.0`)
See a similar situation for Java that was fixed in this PR: https://github.com/pubnub/java/pull/234/files


Chesteer89 (approved):
LGTM, but let's wait for @kleewho opinion

PR #25, opened 2022-05-04
Add entities module

kleewho:


pubnub-entities/src/main/kotlin/com/pubnub/entities/Space.kt
```diff
@@ -0,0 +1,40 @@
+package com.pubnub.entities
+
+import com.pubnub.api.PubNub
+import com.pubnub.api.models.consumer.objects.PNPage
+import com.pubnub.api.models.consumer.objects.PNSortKey
+
+fun PubNub.getSpaces(
```
marcin-cebo, line 7: I like this idea with Extension Methods.

marcin-cebo (approved):

PR #26, opened 2022-05-12
Renamed method parameters and response object types for actions on Sp…

marcin-cebo:
…ace entity.


marcin-cebo:
I will create new branch with fixes.

PR #40, opened 2022-05-30
Add ability to override default release

parfeon:
build(monorepo): add ability to override default release

Add ability to override default release process for monorepo with custom scripts for each product.


kleewho (approved):
Let's see how it goes!

PR #57, opened 2022-06-14
Change package in pubnub-memberships module: entities -> memberships

kleewho:
feat: Now it's possible to receive Membership related events


pubnub-memberships/gradle.properties
```diff
@@ -1,4 +1,4 @@
 POM_ARTIFACT_ID=pubnub-memberships
 POM_NAME=pubnub-memberships
 POM_PACKAGING=jar
-VERSION_NAME=0.1.0
+VERSION_NAME=0.2.0
```
marcin-cebo, line 4: Why we are changing version here. Aren't we changing version in release commit?

kleewho: This is a release commit

PR #59, opened 2022-06-15
UUID to userId in config

marcin-cebo:


pubnub-users/src/test/kotlin/com/pubnub/user/UserExtensionKtTest.kt
```diff
@@ -178,6 +180,22 @@ internal class UserExtensionKtTest {
         verify { pubNub.removeListener(any()) }
     }
 
+    @Test
+    internal fun can_modify_userId() {
+        pubNub.configuration.userId = UserId("modified")
```
kleewho, line 185: How do you know it's not `modified` already? Maybe use `UUID.randomUUID().toString()` instead?

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
```diff
@@ -27,15 +27,23 @@ open class PNConfiguration(
 ) {
 
     init {
-        PubNubUtil.require(uuid.isValid(), PubNubError.UUID_NULL_OR_EMPTY)
+        PubNubUtil.require(uuid.isValid(), PubNubError.USERID_NULL_OR_EMPTY)
     }
 
+    constructor(userId: UserId, enableSubscribeBeta: Boolean = false) : this(userId.value, enableSubscribeBeta)
```
kleewho, line 33: Should we deprecate constructor with `uuid` together with a `uuid` setter?

pubnub-users/src/test/kotlin/com/pubnub/user/UserExtensionKtTest.kt
marcin-cebo: In setUp method there is 
`val pnConfiguration = PNConfiguration(PubNub.generateUUID())`
but for clarity I will modify this test to: 

```
    @Test
    internal fun can_modify_userId() {
        val modifiedUserId = "NewUserId-${UUID.randomUUID()}"
        pubNub.configuration.userId = UserId(modifiedUserId)

        assertEquals(modifiedUserId, pubNub.configuration.userId.value)
    }
```

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
```diff
@@ -27,15 +27,23 @@ open class PNConfiguration(
...
+
     var uuid: String = uuid
         set(value) {
-            PubNubUtil.require(value.isValid(), PubNubError.UUID_NULL_OR_EMPTY)
+            PubNubUtil.require(value.isValid(), PubNubError.USERID_NULL_OR_EMPTY)
```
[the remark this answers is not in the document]
marcin-cebo: Good catch. I will modified it:

```
    var uuid: String = uuid
        set(value) {
            PubNubUtil.require(value.isValid(), PubNubError.UUID_NULL_OR_EMPTY)
            field = value
        }

    var userId: UserId
        get() = UserId(uuid)
        set(value) {
            val userIdValue = value.value
            PubNubUtil.require(userIdValue.isValid(), PubNubError.USERID_NULL_OR_EMPTY)
            uuid = userIdValue
        }
```

pubnub-users/src/main/kotlin/com/pubnub/user/UserExtension.kt
```diff
@@ -82,7 +82,7 @@ private fun toPNSortKey(sort: Collection<ResultSortKey<UserKey>>): Collection<PN
  * @param includeCustom Include respective additional fields in the response.
  */
 fun PubNub.fetchUser(
-    userId: UserId? = UserId(configuration.uuid),
+    userId: UserId? = configuration.userId,
```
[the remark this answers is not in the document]
marcin-cebo: Sure. I will fix that.

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
marcin-cebo: I like this idea. I will add deprecation code.

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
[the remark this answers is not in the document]
marcin-cebo: We could modify UserId class:
data class UserId(val value: String){
    init {
        PubNubUtil.require(value.isValid(), PubNubError.USERID_NULL_OR_EMPTY)
    }
}

kleewho (approved):

PR #60, opened 2022-06-20
Subprojects cleanup

kleewho:
Hopefully everything works

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,63 +20,94 @@ plugins {
     id "com.github.ben-manes.versions" version '0.38.0'
 }
 
-apply plugin: "com.vanniktech.maven.publish"
 
-repositories {
-    mavenCentral()
-    jcenter()
-}
 
-testSets {
-    integrationTest
-}
 
-sourceSets {
-    test {
-        resources {
-            srcDir 'src/testFixtures/resources'
+allprojects {
+    apply plugin: "com.vanniktech.maven.publish"
+
+    if (project.name != "pubnub-kotlin-bom") {
+        apply plugin: 'org.unbroken-dome.test-sets'
+        apply plugin: 'org.jetbrains.kotlin.jvm'
+        apply plugin: "org.jlleitschuh.gradle.ktlint"
+
+
+        repositories {
+            mavenCentral()
+            jcenter()
+        }
+
+        testSets {
+            integrationTest
+        }
+
+        configurations {
+            allTest
+            testFixturesImplementation { extendsFrom allTest }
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -20,63 +20,94 @@ plugins {
...
-testSets {
-    integrationTest
-}
 
-sourceSets {
-    test {
-        resources {
-            srcDir 'src/testFixtures/resources'
+allprojects {
+    apply plugin: "com.vanniktech.maven.publish"
```
marcin-cebo, line 27: Why this plugin can not be included in "plugins" block?

pubnub-users/build.gradle
```diff
@@ -1,31 +1,4 @@
...
-repositories {
-    mavenCentral()
-}
-
-testSets {
-    integrationTest
-}
-
 dependencies {
     implementation project(":")
```
marcin-cebo, line 2: What about an idea of creating pubnub-core module? Isn't it right time to do this?

build.gradle
```diff
@@ -20,63 +20,94 @@ plugins {
...
+            testFixturesImplementation { extendsFrom allTest }
+            integrationTestImplementation { extendsFrom allTest }
+            testImplementation { extendsFrom allTest }
+        }
+
+
+        test {
+            failFast = true
+            exclude '**/contract/*.class'
+            //useJUnitPlatform()
```
marcin-cebo, line 55: What value this comments brings?

pubnub-users/build.gradle
kleewho: I don't know. Is it the right time to do?

build.gradle
kleewho: It wasn't like that because this plugin is not in default gradle plugin repository. But I've added it and it should work just like the other plugins. Also `plugins {}` is not supported in `allprojects` or `subprojects` so I have to still use old `apply plugin`

build.gradle
kleewho: No value. I'll remove it

pubnub-users/build.gradle
marcin-cebo: My understanding is that this change will not be visible to clients. They would configure include our SDK the same way as today. In such case I think that this might be good right time. We can take this off line.

build.gradle
marcin-cebo: ok

pubnub-users/build.gradle
kleewho: Yes, it would not change a thing from the users perspective

PR #61, opened 2022-06-22
Make uuid and userId thread safe.

marcin-cebo:


src/main/kotlin/com/pubnub/api/PNConfiguration.kt
```diff
@@ -36,20 +36,30 @@ open class PNConfiguration(
     )
     constructor(uuid: String, enableSubscribeBeta: Boolean = false) : this(UserId(uuid), enableSubscribeBeta)
 
+    @Volatile
+    private var uuidThreadSafe: String = userId.value
```
kleewho, line 40: Is it really necessary to have another field for that? Wouldn't marking `uuid` enough?

marcin-cebo: I think this is not necessary to have another field. I will use uuid.

kleewho (approved):

PR #62, opened 2022-06-24
Introduced new GrantToken method with no channel groups.

marcin-cebo:
feat: Introduced new GrantToken method with no channel groups.

### DIFF
--- a/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
+++ b/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
@@ -5,8 +5,8 @@ import com.pubnub.api.SpaceId
 import com.pubnub.api.UserId
 import com.pubnub.api.enums.PNLogVerbosity
 import com.pubnub.api.integration.BaseIntegrationTest
-import com.pubnub.api.models.consumer.access_manager.sum.SpaceIdGrant
-import com.pubnub.api.models.consumer.access_manager.sum.UserIdGrant
+import com.pubnub.api.models.consumer.access_manager.sum.SpacePermissions
+import com.pubnub.api.models.consumer.access_manager.sum.UserPermissions
 import com.pubnub.api.models.consumer.access_manager.v3.ChannelGrant
 import com.pubnub.api.models.consumer.access_manager.v3.ChannelGroupGrant
 import com.pubnub.api.models.consumer.access_manager.v3.PNToken.PNResourcePermissions
@@ -31,13 +31,13 @@ class GrantTokenIntegrationTest : BaseIntegrationTest() {
         val grantTokenEndpoint = pubNubUnderTest.grantToken(
             ttl = expectedTTL,
             authorizedUserId = expectedAuthorizedUserId,
-            spaceIds = listOf(
-                SpaceIdGrant.name(spaceId = SpaceId(expectedSpaceIdValue), read = true, delete = true),
-                SpaceIdGrant.pattern(pattern = expectedSpaceIdPattern, write = true, manage = true)
+            spacesPermissions = listOf(
+                SpacePermissions.name(spaceId = SpaceId(expectedSpaceIdValue), read = true, delete = true),
+                SpacePermissions.pattern(pattern = expectedSpaceIdPattern, write = true, manage = true)
             ),
-            userIds = listOf(
-                UserIdGrant.id(userId = UserId(expectedUserIdValue), delete = true),
-                UserIdGrant.pattern(pattern = expectedUserIdPattern, update = true)
+            usersPermissions = listOf(
+                UserPermissions.id(userId = UserId(expectedUserIdValue), delete = true),
+                UserPermissions.pattern(pattern = expectedUserIdPattern, update = true)
             )
         )
 
--- a/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
+++ b/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
@@ -31,7 +31,7 @@ class GrantTokenIntegrationTest : BaseIntegrationTest() {
         val grantTokenEndpoint = pubNubUnderTest.grantToken(
             ttl = expectedTTL,
             authorizedUserId = expectedAuthorizedUserId,
-            spaces = listOf(
+            spaceIds = listOf(
                 SpaceIdGrant.name(spaceId = SpaceId(expectedSpaceIdValue), read = true, delete = true),
                 SpaceIdGrant.pattern(pattern = expectedSpaceIdPattern, write = true, manage = true)
             ),
--- a/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
+++ b/src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
@@ -1,16 +1,70 @@
 package com.pubnub.api.integration.pam
 
 import com.pubnub.api.PubNub
+import com.pubnub.api.SpaceId
+import com.pubnub.api.UserId
 import com.pubnub.api.enums.PNLogVerbosity
 import com.pubnub.api.integration.BaseIntegrationTest
+import com.pubnub.api.models.consumer.access_manager.sum.SpaceIdGrant
+import com.pubnub.api.models.consumer.access_manager.sum.UserIdGrant
 import com.pubnub.api.models.consumer.access_manager.v3.ChannelGrant
 import com.pubnub.api.models.consumer.access_manager.v3.ChannelGroupGrant
 import com.pubnub.api.models.consumer.access_manager.v3.PNToken.PNResourcePermissions
-import org.junit.Assert
+import org.junit.Assert.assertEquals
 import org.junit.Test
 
 class GrantTokenIntegrationTest : BaseIntegrationTest() {
     private val pubNubUnderTest: PubNub = server
+
+    @Test
+    fun happyPath_SUM() {
+        // given
+        pubNubUnderTest.configuration.logVerbosity = PNLogVerbosity.BODY
+        val expectedTTL = 1337
+        val expectedAuthorizedUserId = UserId("authorizedUser01")
+        val expectedSpaceIdValue = "mySpace01"
+        val expectedSpaceIdPattern = "mySpace.*"
+        val expectedUserIdValue = "myUser01"
+        val expectedUserIdPattern = "myUser.*"
+
+        // when
+        val grantTokenEndpoint = pubNubUnderTest.grantToken(
+            ttl = expectedTTL,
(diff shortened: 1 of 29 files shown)


src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
```diff
@@ -1,17 +1,71 @@
...
+            userIds = listOf(
+                UserIdGrant.id(userId = UserId(expectedUserIdValue), delete = true),
+                UserIdGrant.pattern(pattern = expectedUserIdPattern, update = true)
+            )
+        )
+
+        val token = grantTokenEndpoint.sync()!!.token
+
+        // then
+        val (_, _, ttl, _, resources, patterns) = pubNubUnderTest.parseToken(token)
```
kleewho, line 47: oh... niceeee! It seems you're getting used to kotlin

src/main/kotlin/com/pubnub/api/PubNub.kt
```diff
@@ -949,6 +953,40 @@ class PubNub(val configuration: PNConfiguration) {
...
+     * @param authorizedUserId Single userId which is authorized to use the token to make API requests to PubNub
+     * @param spaces List of all channel grants
+     * @param userIds List of all uuid grants
+     */
+    fun grantToken(
+        ttl: Int,
+        meta: Any? = null,
+        authorizedUserId: UserId? = null,
+        spaces: List<SpaceIdGrant> = emptyList(),
+        userIds: List<UserIdGrant> = emptyList()
```
kleewho, line 977: `spaces` and `usersIds`? Would be probably better to have one way of naming it

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/v3/Grants.kt
```diff
@@ -37,7 +40,19 @@ internal data class PNChannelResourceGrant(
     override val get: Boolean = false,
     override val join: Boolean = false,
     override val update: Boolean = false
-) : PNResourceGrant(), ChannelGrant
+) : PNResourceGrant(), ChannelGrant, SpaceIdGrant {
```
kleewho, line 43: So users will be able to pass `PNChannelResourceGrant` to new `GrantToken`? Is this what we actually want?

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/sum/UserIdGrant.kt
```diff
@@ -0,0 +1,45 @@
...
+            delete = delete,
+            get = get,
+            update = update
+        )
+    }
+}
+
+fun UserIdGrant.toUuidGrants(): UUIDGrant {
+    return when (this) {
+        is PNUUIDResourceGrant -> PNUUIDResourceGrant(userIdGrant = this)
```
kleewho, line 41: What? I have trouble understanding this line. If we have `PNUUIDResourceGrant` then we'll create `PNUUIDResourceGrant` passing this thing as an argument. But why? We already have `PNUUIDResourceGrant`, so what's the point?

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/v3/Grants.kt
```diff
@@ -37,7 +40,19 @@ internal data class PNChannelResourceGrant(
...
+    constructor(spaceIdGrant: SpaceIdGrant) : this(
```
kleewho, line 44: I don't think I like that old classes knows anything about new classes

kleewho (changes requested):

src/integrationTest/kotlin/com/pubnub/api/integration/pam/GrantTokenIntegrationTest.kt
marcin-cebo: :)

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/v3/Grants.kt
marcin-cebo: val grantTokenEndpoint = pubNubUnderTest.grantToken(
            ttl = expectedTTL,
            authorizedUserId = expectedAuthorizedUserId,
            spaceIds = listOf(
                SpaceIdGrant.name(spaceId = SpaceId(expectedSpaceIdValue), read = true, delete = true),
                SpaceIdGrant.pattern(pattern = expectedSpaceIdPattern, write = true, manage = true)
            ),
            userIds = listOf(
                UserIdGrant.id(userId = UserId(expectedUserIdValue), delete = true),
                UserIdGrant.pattern(pattern = expectedUserIdPattern, update = true)
            )
        )

PNChannelResourceGrant is internal  so it seems that it will be difficult for client to use `PNChannelResourceGrant` to new `GrantToken`

Other option would be to introduce new interfaces/classes that will be copy of existing interfaces/classes from Grants.kt. I can do this. Having them there will less work in the future when Channel/UUID related elements will be removed.

src/main/kotlin/com/pubnub/api/PubNub.kt
marcin-cebo: yes, sure.  I forgot to change spaces to spaceIds.

src/main/kotlin/com/pubnub/api/models/consumer/access_manager/v3/Grants.kt
marcin-cebo: ok, I will create set of new classes

src/main/kotlin/com/pubnub/api/PubNub.kt
kleewho: maybe space/user permissions? I mean it's not only id, right?

src/test/kotlin/com/pubnub/api/endpoints/access/GrantTokenTest.kt
```diff
@@ -0,0 +1,105 @@
...
+                    ChannelGrant.name(name = "channel1", read = true, delete = true),
+                    ChannelGrant.pattern(pattern = "channel.*", write = true, manage = true)
+                ),
+                channelGroups = emptyList(),
+                uuids = emptyList()
+            )
+        )
+
+        every { pubnub.grantToken(ttl = any(), authorizedUserId = any(), spaceIds = any()) } returns grantTokenEndpointMock
+        every { grantTokenEndpointMock.doWork(any()) } returns call
```
kleewho, line 87: If it ok if we have a test which doesn't test `doWork` in any way? I always struggle with the value that those tests bring for the cost that we pay. When we disabling some part of our code and use mocks I'm struggling even more.

src/test/kotlin/com/pubnub/api/endpoints/access/GrantTokenTest.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    @MockK
+    private lateinit var responseFromServer: Response<GrantTokenResponse>
+
+    @MockK
+    private lateinit var rawResponse: okhttp3.Response
+
+    @Test
+    fun can_createGrantTokenSimple() {
+        val expectedTTL = 1337
```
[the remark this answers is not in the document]
marcin-cebo: +#/\[/]x$

src/main/kotlin/com/pubnub/api/endpoints/access/GrantToken.kt
```diff
@@ -39,7 +39,7 @@ class GrantToken(
         }
     }
 
-    override fun doWork(queryParams: HashMap<String, String>): Call<GrantTokenResponse> {
+    public override fun doWork(queryParams: HashMap<String, String>): Call<GrantTokenResponse> {
```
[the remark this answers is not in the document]
marcin-cebo: Based on my experience unit test is group of test that should be most dominating group of test in app.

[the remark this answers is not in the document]
kleewho: I guess these comments are a response to my different comment which I'll bring here for context.

> If it ok if we have a test which doesn't test doWork in any way? I always struggle with the value that those tests bring for the cost that we pay. When we disabling some part of our code and use mocks I'm struggling even more.

I agree that we should have the most unit tests. 

> I didn't find better way to write unit test for grantToken.

Here's different approach which not disables the code in `doWork`. Still relies on mock heavily which I don't like

```
    fun fakeCall() = object : Call<GrantTokenResponse> {
        override fun clone(): Call<GrantTokenResponse> {
            return this
        }

        override fun execute(): Response<GrantTokenResponse> {
            return Response.success(GrantTokenResponse(GrantTokenData("aaaa")))
        }

        override fun isExecuted(): Boolean {
            TODO("Not yet implemented")
        }

        override fun cancel() {
            TODO("Not yet implemented")
        }

        override fun isCanceled(): Boolean {
            TODO("Not yet implemented")
        }

        override fun request(): Request {
            TODO("Not yet implemented")
        }

        override fun timeout(): Timeout {
            TODO("Not yet implemented")
        }

        override fun enqueue(callback: Callback<GrantTokenResponse>) {
            TODO("Not yet implemented")
        }
    }

    @Test
    fun can_beAnExample() {
        val expectedTTL = 1337
        val authorizedUserId = UserId("authorizedUserId")
        val expectedToken = "token_value"
        val grantTokenResult = PNGrantTokenResult(token = expectedToken)

        val retrofitManager = mockk<RetrofitManager>(relaxed = true)
        val accessManagerService = mockk<AccessManagerService>(relaxed = true)
        val capturedBodies = mutableListOf<Any>()
        val capturedParams = mutableListOf<Map<String, String>>()


        every { pubnub.retrofitManager } returns retrofitManager
        every { retrofitManager.accessManagerService } returns accessManagerService
        every { accessManagerService.grantToken(any(), capture(capturedBodies), capture(capturedParams)) } returns fakeCall()

        val result = pubnub.grantToken(
            ttl = expectedTTL,
            authorizedUserId = authorizedUserId,
            spaces = listOf(SpaceIdGrant.name(spaceId = SpaceId("mySpaceId"), read = true, delete = true))
        ).sync()

        val capturedBody = capturedBodies[0]
        //asserts on body
        val params = capturedParams[0]
        //asserts on params
    }
```

src/main/kotlin/com/pubnub/api/PubNub.kt
marcin-cebo: I like UserPermissions and SpacePermissions. I will change it.

src/test/kotlin/com/pubnub/api/endpoints/access/GrantTokenTest.kt
marcin-cebo: ok

kleewho (approved):

marcin-cebo:
@client-engineering-bot  release kotlin as v7.3.0

marcin-cebo:
@client-engineering-bot  release kotlin as v7.3.0

marcin-cebo:
@client-engineering-bot release kotlin as v7.3.0

PR #84, opened 2022-10-25
Address warnings in the build

kleewho:
* To address deprecation in most places not deprecated has been used. In those specific areas where we actually want to verify that our deprecated methods are working I added SUPPRESS annotations.
* ktlint debug has been disabled to reduce the amount of information produced
* .editorconfig file used the suggested by ktlint value


pubnub-users/src/integrationTest/kotlin/com/pubnub/user/UserIntegTest.kt
```diff
@@ -39,7 +38,8 @@ class UserIntegTest {
 
     @BeforeEach
     fun setUp() {
-        val config = PNConfiguration(UserId("kotlin")).apply {
+        UserId("kotlin")
```
marcin-cebo, line 41: What is this line for?

src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
```diff
@@ -103,8 +103,8 @@ class FetchMessages internal constructor(
             effectiveMax(page.limit, includeMessageActions, channels.size).toString()
         queryParams["include_uuid"] = includeUUID.toString()
 
-        page.start?.run { queryParams["start"] = this.toString().toLowerCase(Locale.US) }
-        page.end?.run { queryParams["end"] = this.toString().toLowerCase(Locale.US) }
+        page.start?.run { queryParams["start"] = this.toString().lowercase(Locale.US) }
```
marcin-cebo, line 106: Why in some places we have:
Locale.getDefault() and here Locale.US ?

marcin-cebo (changes requested):

src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
kleewho: That's a very good question and I don't know. It's not an issue for a start or an end, but I guess we should have the same locales in the code

pubnub-users/src/integrationTest/kotlin/com/pubnub/user/UserIntegTest.kt
kleewho: That's a leftover after failed attempt of mass fix

pubnub-spaces/src/integrationTest/kotlin/com/pubnub/space/SpaceIntegTest.kt
```diff
@@ -38,7 +37,8 @@ class SpaceIntegTest {
 
     @BeforeEach
     fun setUp() {
-        val config = PNConfiguration(UserId("kotlin")).apply {
+        UserId("kotlin")
```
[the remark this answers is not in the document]
kleewho: That's a leftover after failed attempt of mass fix

marcin-cebo (approved):

PR #90, opened 2022-12-09
Add messageType and spaceId to publish

kleewho:


src/test/kotlin/com/pubnub/contract/step/KeysetStep.kt
```diff
@@ -38,4 +38,15 @@ class KeysetStep(private val world: World) {
             subscribeKey = ContractTestConfig.subKey
         }
     }
+
+    @Given("the demo keyset")
+    fun the_demo_keyset() {
+        MatcherAssert.assertThat(ContractTestConfig.pubKey, Matchers.notNullValue())
```
marcin-cebo, line 44: What about replacing body with:
i_have_a_keyset_with_access_manager_enabled_without_secret_key()
?

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
```diff
@@ -15,17 +15,23 @@ import com.pubnub.api.CommonUtils.assertPnException
 import com.pubnub.api.CommonUtils.failTest
 import com.pubnub.api.PubNubError
 import com.pubnub.api.PubNubException
+import com.pubnub.api.SpaceId
```
marcin-cebo, line 18: What do you think  about adding/modifying integration tests to cover new fields?

src/main/kotlin/com/pubnub/api/PubNub.kt
```diff
@@ -191,7 +192,9 @@ class PubNub(val configuration: PNConfiguration) {
         shouldStore: Boolean? = null,
         usePost: Boolean = false,
         replicate: Boolean = true,
-        ttl: Int? = null
+        ttl: Int? = null,
+        spaceId: SpaceId? = null,
```
marcin-cebo, line 196: Will those parameters be not accessible by client?

I am thinking that sooner o latter we will create branch related to messageType and spaceId. Maybe we can create it now and store in it all changes related to this work?

src/main/kotlin/com/pubnub/api/models/consumer/MessageType.kt
```diff
@@ -0,0 +1,30 @@
+package com.pubnub.api.models.consumer
+
+import com.pubnub.api.PubNubException
+
+sealed interface MessageType {
+    companion object {
+        operator fun invoke(value: String): MessageType {
+            return UserDefined(value)
+        }
+        internal fun of(value: Int?): MessageType = when (value) {
```
marcin-cebo, line 10: How this "of" method will be used?
Can we have a test that demonstrate this?

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
```diff
@@ -562,4 +534,84 @@ class PublishTest : BaseTest() {
         assertEquals("0", requests[0].queryParameter("store").firstValue())
         assertFalse(requests[0].queryParameter("ttl").isPresent)
     }
+
+    @Test
+    fun testSpaceIdQueryParamIsPassedInPublish() {
+        val spaceIdValue = "thisIsSpaceId"
+
+        stubFor(
+            get(urlPathEqualTo("/publish/myPublishKey/mySubscribeKey/0/coolChannel/0/%22hi%22"))
```
marcin-cebo, line 543: What about extracting "coolChannel" to variable. Then the variable can be used here and in publish call:
channel = "coolChannel",
?

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
```diff
@@ -562,4 +534,84 @@ class PublishTest : BaseTest() {
...
+        assertEquals(1, requests.size)
+        assertThat(requests[0].queryParams, not(hasKey(Publish.SPACE_ID_QUERY_PARAM)))
+    }
+
+    @Test
+    fun testMessageTypeQueryParamIsPassedInPublish() {
+        val messageTypeValue = "thisIsSpaceId"
+
+        stubFor(
+            get(urlPathEqualTo("/publish/myPublishKey/mySubscribeKey/0/coolChannel/0/%22hi%22"))
```
marcin-cebo, line 583: What about extracting "coolChannel" to variable. Then the variable can be used here and in publish call:
channel = "coolChannel",
?

marcin-cebo (changes requested):

src/main/kotlin/com/pubnub/api/models/consumer/MessageType.kt
kleewho: `MessageType.of(3)`

src/main/kotlin/com/pubnub/api/PubNub.kt
kleewho: They will be available for users. Not sure what branch should I create. I guess what's missing is the kdoc for them

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
kleewho: Sure. Why not

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
kleewho: Sure. Why not

src/main/kotlin/com/pubnub/api/models/consumer/MessageType.kt
kleewho: Do we need such a test? I can write it, but not sure about the added value

src/test/kotlin/com/pubnub/contract/step/KeysetStep.kt
kleewho: I don't know. Is it the same because it's the nature of the requirement or is it the same by accident?

kleewho: I can do whatever. I think those steps are mostly not interesting for SDK at all

src/test/kotlin/com/pubnub/api/legacy/endpoints/pubsub/PublishTest.kt
kleewho: Added one case

src/test/kotlin/com/pubnub/contract/step/KeysetStep.kt
marcin-cebo: I see

marcin-cebo (approved):
Looks good to me.

PR #93, opened 2022-12-21
Fix command handler workflow

parfeon:
fix(github-actions): fix command handler workflow

Fix how command handler workflow verify whether comment is command for bot or not.


.github/workflows/run-validations.yml
```diff
@@ -1,24 +1,49 @@
...
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Validate clean build
+        if: ${{ github.event_name != 'pull_request' && github.ref == 'refs/heads/master' }}
+        run: |
+          ./gradlew build
```
kleewho, line 49: Why do we need this here? Why can't it be the part of tests? It's a waste of time IMHO. It'll do the tests and a little bit more

.github/workflows/validate-yml.js
```diff
@@ -1,94 +0,0 @@
...
-            return [sdkURLs, featureMappingURL, pubnubYAMLSchemaURL];
-        } else {
-            console.log(chalk.red("response locations null"));
-            return null;
-        }
-    } else {
-        console.log(chalk.red("response body null"));
-        return null;
-    }
-}
```
kleewho, line 94: AWESOME!

.github/workflows/run-validations.yml
parfeon: Is there anything, that `gradle` will be able to reuse after tests with `gradlew build` command?
I'm trying to separate them contextually, This is workflow, in which various checks will be performed. In Swift we have checks on whether it compatible with dependency managers and they can handle it.

.github/workflows/commands-handler.yml
```diff
@@ -3,27 +3,42 @@ name: Commands processor
...
 jobs:
   process:
     name: Process command
-    if: ${{ github.event.issue.pull_request && endsWith(github.repository, '-private') != true && startsWith(github.event.comment.body, '@client-engineering-bot ') }}
+    if: ${{ github.event.issue.pull_request && endsWith(github.repository, '-private') != true
     runs-on: ubuntu-latest
     steps:
+      - name: Check referred user
+        id: user-check
+        shell: bash
```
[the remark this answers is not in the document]
parfeon: It won't make things worse, but yeah.. I copy pasted it accidentally :/
Just waiting from you more info on validation.

.github/workflows/run-validations.yml
parfeon: @kleewho  Actually, there is no time will be wasted this way,. because it is in separate job, which runs in parallel, compared to when it was one of steps in `run-tests.yml` workflow.

kleewho: > Actually, there is no time will be wasted this way,. because it is in separate job, which runs in parallel,

Ok. CPU cycles wasted :D 

> Is there anything, that gradle will be able to reuse after tests 

There's probably a lot that could be reused but only if tests had chance to run before and put the results into the cache. Because they'll run in parallel it won't happen

parfeon: Ok, how critical it to be in same workflow as tests? 
Right now there is a win in overall duration of checks, because all will start in parallel jobs. No win in context of time in total (if we count each job and sum time for them).

I thought it will be great to have it contextually separated, because it is not really a test, but verification of whether build can be made or not.

If you think, that it should be in tests workflow, I'll move it back there.

kleewho: Sure. We can have this contextually separated

kleewho (approved):

PR #96, opened 2023-01-02
Use atomic to have sequence without synchronization on Int

kleewho:
This solves the last warning we had during the compilation time:

> w: pubnub/kotlin/src/main/kotlin/com/pubnub/api/managers/PublishSequenceManager.kt: (8, 9): Synchronizing by Int is forbidden

`getAndUpdate` is supported on Android thanks to desugaring https://developer.android.com/studio/write/java8-support-table


Chesteer89 (approved):

marcin-cebo (commented):
Nice

PR #97, opened 2023-01-04
Subscribe mt si contract test

kleewho:
I would like to change how those steps are organized a little. New approached is already present in this PR. Other should come afterwards. What was wrong IMHO is that I named all steps regarding of the feature the same e.g. `WhenSteps`, `GivenSteps` which made it harder to navigate the code. I still feel pretty strongly about grouping them per feature, but the feature name should be in the name file as well. Also I don't think the packages `step`, `state`, etc. are so important. We can of course talk about it and if necessary I can adjust this change back to the old structure

### DIFF
--- a/src/main/kotlin/com/pubnub/api/models/consumer/presence/PNHereNow.kt
+++ b/src/main/kotlin/com/pubnub/api/models/consumer/presence/PNHereNow.kt
@@ -24,10 +24,10 @@ class PNHereNowResult internal constructor(
  * @property occupancy Total number of UUIDs currently in the channel.
  * @property occupants List of [PNHereNowOccupantData] (users) currently in the channel.
  */
-class PNHereNowChannelData internal constructor(
+data class PNHereNowChannelData internal constructor(
     val channelName: String,
     val occupancy: Int,
-    var occupants: List<PNHereNowOccupantData> = emptyList()
+    val occupants: List<PNHereNowOccupantData> = emptyList()
 )
 
 /**
--- a/src/main/kotlin/com/pubnub/api/services/PresenceService.kt
+++ b/src/main/kotlin/com/pubnub/api/services/PresenceService.kt
@@ -31,14 +31,14 @@ internal interface PresenceService {
         @QueryMap options: Map<String, String>
     ): Call<Envelope<WhereNowPayload>>
 
-    @GET("v2/presence/sub_key/{subKey}/channel/{channel}")
+    @GET("v2/presence/sub-key/{subKey}/channel/{channel}")
     fun hereNow(
         @Path("subKey") subKey: String,
         @Path("channel") channel: String,
         @QueryMap options: Map<String, String>
     ): Call<Envelope<JsonElement>>
 
-    @GET("v2/presence/sub_key/{subKey}")
+    @GET("v2/presence/sub-key/{subKey}")
     fun globalHereNow(
         @Path("subKey") subKey: String,
         @QueryMap options: Map<String, String>
--- a/src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt
+++ b/src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt
@@ -28,7 +28,7 @@ class HereNowEndpointTest : BaseTest() {
     @Test
     fun testMultipleChannelStateSync() {
         stubFor(
-            get(urlPathEqualTo("/v2/presence/sub_key/mySubscribeKey/channel/ch1,ch2")).willReturn(
+            get(urlPathEqualTo("/v2/presence/sub-key/mySubscribeKey/channel/ch1,ch2")).willReturn(
                 aResponse().withBody(
                     """
                     {
@@ -105,7 +105,7 @@ class HereNowEndpointTest : BaseTest() {
     @Test
     fun testMultipleChannelSync() {
         stubFor(
-            get(urlPathEqualTo("/v2/presence/sub_key/mySubscribeKey/channel/ch1,ch2"))
+            get(urlPathEqualTo("/v2/presence/sub-key/mySubscribeKey/channel/ch1,ch2"))
                 .willReturn(
                     aResponse().withBody(
                         """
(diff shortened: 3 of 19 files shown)


marcin-cebo (approved):

src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt
```diff
@@ -105,7 +105,7 @@ class HereNowEndpointTest : BaseTest() {
     @Test
     fun testMultipleChannelSync() {
         stubFor(
-            get(urlPathEqualTo("/v2/presence/sub_key/mySubscribeKey/channel/ch1,ch2"))
+            get(urlPathEqualTo("/v2/presence/sub-key/mySubscribeKey/channel/ch1,ch2"))
```
marcin-cebo, line 108: Why this change?
In openApi there is "sub_key"

src/main/kotlin/com/pubnub/api/services/PresenceService.kt
```diff
@@ -31,14 +31,14 @@ internal interface PresenceService {
...
-    @GET("v2/presence/sub_key/{subKey}/channel/{channel}")
+    @GET("v2/presence/sub-key/{subKey}/channel/{channel}")
     fun hereNow(
         @Path("subKey") subKey: String,
         @Path("channel") channel: String,
         @QueryMap options: Map<String, String>
     ): Call<Envelope<JsonElement>>
 
-    @GET("v2/presence/sub_key/{subKey}")
+    @GET("v2/presence/sub-key/{subKey}")
```
marcin-cebo, line 41: I surprised. Was this change somehow announced?

src/main/kotlin/com/pubnub/api/models/consumer/presence/PNHereNow.kt
```diff
@@ -24,10 +24,10 @@ class PNHereNowResult internal constructor(
  * @property occupancy Total number of UUIDs currently in the channel.
  * @property occupants List of [PNHereNowOccupantData] (users) currently in the channel.
  */
-class PNHereNowChannelData internal constructor(
+data class PNHereNowChannelData internal constructor(
     val channelName: String,
     val occupancy: Int,
-    var occupants: List<PNHereNowOccupantData> = emptyList()
+    val occupants: List<PNHereNowOccupantData> = emptyList()
```
Chesteer89, line 30: 👍

src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt
marcin-cebo: I see PR from Serhii. I am wondering why openAPI doesn't reflect those changes?

src/main/kotlin/com/pubnub/api/services/PresenceService.kt
marcin-cebo: I see PR from Serhii. I am wondering why openAPI doesn't reflect those changes?

Chesteer89: Can we make sure it works well? Additionally, can we check all the `sub-key` usage? We found that `sub-key` and `sub_key` are mixed in various SDKs

src/test/kotlin/com/pubnub/contract/step/ThenSteps.kt
```diff
@@ -24,7 +24,10 @@ class ThenSteps(private val world: World) {
         assertNull(world.pnException)
     }
 
-    @Then("I receive a successful response")
+    @Then.Thens(
```
Chesteer89, line 27: Why `@Then.Thens`?

src/main/kotlin/com/pubnub/api/services/PresenceService.kt
Chesteer89: Please make sure that all the SDKs are implementing it the same way. We're using it in Chat Team as well (we'll have to fix the contracts)

marcin-cebo (approved):
Nice

marcin-cebo (approved):

src/test/kotlin/com/pubnub/contract/history/HistorySteps.kt
```diff
@@ -6,7 +6,6 @@ import com.pubnub.contract.state.World
 import io.cucumber.java.en.Then
```
marcin-cebo, line 6: In @Then("history response contains messages with {string} and {string} message types")
you could go with like:
  ```
      val messageTypeValuesInAllMessagesInAllChannels =
            fetchMessagesState.pnFetchMessagesResult?.getAllFetchMessageItemsFromAllChannels()
                ?.map { it.messageType.value }

        val uniqueMessageTypeValuesInAllMessagesInAllChannels: HashSet<String> = HashSet(messageTypeValuesInAllMessagesInAllChannels)

        assertThat(
            uniqueMessageTypeValuesInAllMessagesInAllChannels,
            containsInAnyOrder(messageTypeOfFirstMessage, messageTypeOfSecondMessage)
        )
```

**Review by marcin-cebo — APPROVED** [review]

File: src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt (hunk 1 above)
Comment (kleewho) [reply src/test/kotlin/com/pubnub/api/legacy/endpoints/presence/HereNowEndpointTest.kt:108]: AFAIK our server doesn't care, and we have a mess on SDK side. Would be nice to have just one way of referencing it. This was chosen because at the time of change Serhii was in the office and on ObjectiveC everywhere is `-`. I don't care what character we will use, as long as it's consistent

File: src/main/kotlin/com/pubnub/api/services/PresenceService.kt (hunk 2 above)
Comment (kleewho) [reply src/main/kotlin/com/pubnub/api/services/PresenceService.kt:41]: @Chesteer89 sure. I'll make sure that we have consistent paths at least across SDKS used in chat components before merging this

File: src/main/kotlin/com/pubnub/api/models/consumer/presence/PNHereNow.kt (hunk 3 above)
Comment (kleewho) [reply src/main/kotlin/com/pubnub/api/models/consumer/presence/PNHereNow.kt:30]: I know ;)

File: src/test/kotlin/com/pubnub/contract/history/HistorySteps.kt (hunk 5 above)
Comment (kleewho) [reply src/test/kotlin/com/pubnub/contract/history/HistorySteps.kt:6]: Yeaaah. Sure. In Kotlin it would be better like that:

```
fetchMessagesState.pnFetchMessagesResult?.getAllFetchMessageItemsFromAllChannels()
              ?.map { it.messageType.value }.distinct()
```

or if you really insist on set than like this:

```
fetchMessagesState.pnFetchMessagesResult?.getAllFetchMessageItemsFromAllChannels()
              ?.map { it.messageType.value }.toSet()
```

by I would prefer the first approach because it's more explicit.

PR #98, opened 2023-01-16
Add support for messageType and spaceId to fetchMessages

kleewho:
The reason for HistoryMessageType sealed interface is that users will learn from the type system what possible types we can return:

```

    fun fetchMessageItem(): PNFetchMessageItem {
        TODO()
    }

    fun example() {
       when (fetchMessageItem().messageType) { //this when is already exhaustive
           MessageType.File -> TODO()
           MessageType.Message -> TODO()
           is MessageType.UserDefined -> TODO()
           null -> TODO()
       }
    }
```

### DIFF
--- a/src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
+++ b/src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
@@ -98,7 +98,7 @@ class FetchMessages internal constructor(
                 val newActions = if (includeMessageActions) serverMessageItem.actions ?: mapOf() else serverMessageItem.actions
                 val messageType = if (includeMessageType) {
                     serverMessageItem.userDefinedMessageType?.let { MessageType.UserDefined(it) }
-                        ?: serverMessageItem.integerMessageType.let { HistoryMessageType.of(it) }
+                        ?: serverMessageItem.pnMessageType.let { HistoryMessageType.of(it) }
                 } else {
                     null
                 }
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
```diff
@@ -107,5 +131,10 @@ class FetchMessages internal constructor(
         page.end?.run { queryParams["end"] = this.toString().lowercase(Locale.US) }
 
         if (includeMeta) queryParams["include_meta"] = includeMeta.toString()
```
marcin-cebo, line 133: In this case:
includeMeta.toString()
includeMessageType.toString()
includeSpaceId.toString()
are always true.

What do you think about changing those values to "true" ?

src/main/kotlin/com/pubnub/api/models/server/history/ServerFetchMessageItem.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.pubnub.api.models.consumer.history.Action
+
+data class ServerFetchMessageItem(
+    val uuid: String?,
+    val message: JsonElement,
+    val meta: JsonElement?,
+    val timetoken: Long,
+    val actions: Map<String, Map<String, List<Action>>>? = null,
+    @SerializedName("message_type")
+    val integerMessageType: Int? = null,
```
marcin-cebo, line 14: What about changing it to pnMessageType  or pnMessageTypeInt for consistence with other usages e.g. in SubscribeMessage there is pnMessageTypeInt. ?

src/main/kotlin/com/pubnub/api/models/server/history/ServerFetchMessageItem.kt
```diff
@@ -0,0 +1,18 @@
+package com.pubnub.api.models.server.history
+
+import com.google.gson.JsonElement
+import com.google.gson.annotations.SerializedName
+import com.pubnub.api.models.consumer.history.Action
+
+data class ServerFetchMessageItem(
```
marcin-cebo, line 7: I like this class.

marcin-cebo (changes requested):

src/main/kotlin/com/pubnub/api/models/server/history/ServerFetchMessageItem.kt
kleewho: Yes. I missed this one :(

src/main/kotlin/com/pubnub/api/endpoints/FetchMessages.kt
kleewho: Sure. I'll do it

marcin-cebo (approved):

PR #100, opened 2023-01-30
Add step implementation required by signal-to-space.feature

kleewho:


src/test/kotlin/com/pubnub/contract/signal/SignalSteps.kt
```diff
@@ -0,0 +1,26 @@
+package com.pubnub.contract.signal
+
+import com.pubnub.api.PubNubException
+import com.pubnub.api.SpaceId
+import com.pubnub.api.models.consumer.MessageType
+import com.pubnub.contract.state.World
+import io.cucumber.java.en.When
+
+class SignalSteps(private val world: World) {
+    @When("I send a signal with {spaceId} space id and {string} message type")
```
marcin-cebo, line 10: Changing from string to spaceId. Is it working? Do tests pass?

kleewho: yes. It's thanks to this thing:

```
@ParameterType("'(.*)'")
fun spaceId(stringId: String): SpaceId {
    return SpaceId(stringId)
}
```

marcin-cebo: ok, I didn't see this. And why there is no similar approach for messageType?

kleewho: Because messageType is not that straightforward. Let's consider step:

>     And history response contains messages with 'message' and 'vc-message' message types

Here `'message'` most probably mean that it's an internal message type which in kotlin is:

```
object Message : PNMessageType, HistoryMessageType
```

Based only on this information what kind of string we have I only can guess which exactly type to create. If I create wrong type the message types won't be equal and I'll have failed tests. That's why I rather compare the values only in this case. That's the longer story. 

Short story is that at the beginning I didn't bother to create it, and just by sheer luck it happened that I see some problems with it after implementing subscribe steps.

marcin-cebo (approved):

PR #101, opened 2023-02-02
Set properly encrypted file size when uploading

kleewho:
fix: Uploading an encrypted file correctly sets the content size in the multipart body.


src/test/kotlin/com/pubnub/api/legacy/endpoints/files/SendFileTest.kt
```diff
@@ -63,20 +51,61 @@ class SendFileTest : TestsWithFiles {
...
+            sendFile(channel, fileName(), inputStream).sync()
         }
 
         // then
         Assert.assertEquals(expectedResponse, result)
     }
 
     @Test
-    @Throws(InterruptedException::class, IOException::class)
+    fun spaceIdAndMessageTypeArePassedToPublishMessageFile() {
```
marcin-cebo, line 63: Are you sure that test does what it promises? Is this possible to pass spaceId and messageType to publishMessageType?

marcin-cebo (changes requested):

src/test/kotlin/com/pubnub/api/legacy/endpoints/files/SendFileTest.kt
kleewho: Nope. This shouldn't be there. It's a spoiler from other PR that will come soon ;)

marcin-cebo: ok

marcin-cebo (approved):

kleewho:
@pubnub-release-bot release kotlin as v7.4.1

PR #102, opened 2023-02-02
Support for SpaceId and MessageType in files

kleewho:


src/integrationTest/kotlin/com/pubnub/api/integration/FilesIntegrationTest.kt
```diff
@@ -1,151 +1,75 @@
...
-        uploadListDownloadDelete(true)
-    }
 
-    @Test
-    fun uploadListDownloadDeleteWithoutCipher() {
-        uploadListDownloadDelete(false)
-    }
+    @ParameterizedTest
+    @ValueSource(strings = ["enigma", ""])
+    fun sendListDownloadDelete(cipher: String) {
```
marcin-cebo, line 19: I like code in this test. IMHO is sophisticated but still bit difficult for me to read. Maybe because of limited experience with Kotlin.

src/test/kotlin/com/pubnub/api/legacy/endpoints/files/PublishFileMessageTest.kt
```diff
@@ -0,0 +1,92 @@
...
+                aResponse().withBody("""[1,"Sent","15883272000000000"]""")
+            )
+        )
+    }
+
+    @Test
+    fun testSpaceIdQueryParamIsPassedInPublish() {
+        val spaceIdValue = "thisIsSpaceId"
+
+        pubnub.publishFileMessage(
```
marcin-cebo, line 38: I like this test. So wire mock can be used to mimic REST endpoint response but also to check query parameter value/presence.

marcin-cebo (approved):

src/test/kotlin/com/pubnub/api/legacy/endpoints/files/PublishFileMessageTest.kt
marcin-cebo: Not sure what you mean by "used to" but I also like approach without using WireMock like in FetchMessagesTest in java where we do:
Call<FetchMessagesEnvelope> call = objectUnderTest.doWork(baseParams);

This approach shows more details about FetchMessage class internals.

marcin-cebo (approved):

PR #104, opened 2023-02-08
Enable integration tests

kleewho:
Only junit 5 tests are running for integrationTests. Thanks to that we can slowly add them one by one after cleaning them up


src/test/kotlin/com/pubnub/api/CommonUtils.kt
```diff
@@ -0,0 +1,101 @@
+package com.pubnub.api
+
+import com.github.tomakehurst.wiremock.client.WireMock
```
marcin-cebo, line 3: There is CommonUtils.kt file located here:
src/integrationTest/kotlin/com/pubnub/api/CommonUtils.kt
that has very similar content. Having two of them is desired?

marcin-cebo (approved):

src/test/kotlin/com/pubnub/api/CommonUtils.kt
kleewho: It's necessary for the tests to compile. Those two sources can't share the code between them, and because I've removed the fixtures plugin we need to have the copy in both places. Eventually when we reduce the number of integrationTests we might not need it

PR #110, opened 2023-04-06
Do not cancel subscribe calls on set state

kleewho:
fix: Do not cancel subscribe call on set state


src/main/kotlin/com/pubnub/api/managers/SubscriptionManager.kt
```diff
@@ -107,7 +107,7 @@ class SubscriptionManager(val pubnub: PubNub, private val subscriptionState: Sta
     }
 
     internal fun adaptStateBuilder(stateOperation: StateOperation) {
-        reconnect(stateOperation)
+        subscriptionState.handleOperation(stateOperation)
```
kleewho, line 110: This is the change

marcin-cebo (approved):

kleewho:
@pubnub-release-bot release kotlin

PR #113, opened 2023-04-19
Add messageType support to fetchMessages

kleewho:
feat: It's now possible to return messageType in PNFetchMessageItem from fetchMessages method. Possible values are either Message or File


are (approved):

kleewho:
@pubnub-release-bot release kotlin

PR #121, opened 2023-05-05
For each effect create separate class.

marcin-cebo:
Make SubscribeManagedEffectFactory less responsible.


src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffect.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.pubnub.api.PubNubException
+import com.pubnub.api.endpoints.remoteaction.RemoteAction
+import com.pubnub.api.eventengine.EventDeliver
+import com.pubnub.api.eventengine.ManagedEffect
+import com.pubnub.api.subscribe.eventengine.event.Event
+import com.pubnub.api.subscribe.eventengine.event.SubscriptionCursor
+
+class HandshakeEffect(
+    private val handshakeProvider: HandshakeProvider,
+    private val eventDeliver: EventDeliver,
```
marcin-cebo, line 12: I would rename eventDeliver to:
interface EventQueue {
    fun add(event: Event)
    fun take() : Event
}

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/DelayedManagedEffect.kt
```diff
@@ -1,46 +0,0 @@
-package com.pubnub.api.subscribe.eventengine.effect
-
-import com.pubnub.api.eventengine.ManagedEffect
-import java.time.Duration
-import java.util.concurrent.ScheduledExecutorService
-import java.util.concurrent.ScheduledFuture
-import java.util.concurrent.TimeUnit
-
-class DelayedManagedEffect(
```
kleewho, line 9: What's the reason for removal?

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffect.kt
```diff
@@ -0,0 +1,42 @@
...
+                    eventDeliver.passEventForHandling(Event.HandshakeSuccess(result!!))
+                }
+            } finally {
+                completionBlock()
+            }
+        }
+    }
+
+    override fun cancel() {
+        val handshakeRemoteAction = getHandshakeRemoteAction()
```
kleewho, line 36: This will only work if `getHandshakeRemoteAction` returns same instance. I don't think we should assume this therefore this is a bug. It's not possible to really cancel this effect. Same mistake is present in other classes.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/RemoteActionManagedEffect.kt
```diff
@@ -1,28 +0,0 @@
-package com.pubnub.api.subscribe.eventengine.effect
-
-import com.pubnub.api.endpoints.remoteaction.RemoteAction
-import com.pubnub.api.eventengine.ManagedEffect
-import com.pubnub.api.models.consumer.PNStatus
-
-class RemoteActionManagedEffect<T>(
```
kleewho, line 7: What's the reason for removal? Together with `DelayedManagedEffect` it seems to be powerful enough to fulfill our needs and on the other hand the implementation of both is concise and focused

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun `should deliver HandshakeSuccess event when HandshakeEffect succeeded `() {
+        // given
+        every { handshakeProvider.handshake(any(), any()) } returns successfulRemoteAction(subscriptionCursor)
+        val handshakeEffect = HandshakeEffect(handshakeProvider, eventDeliver, handshakeInvocation)
+
+        // when
+        handshakeEffect.runEffect()
+
+        // then
+        Thread.sleep(50)
```
kleewho, line 32: I don't think we need any of those `Thread.sleep`. There's got to be a better, more reliant and faster way to do the same.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // todo add test to check completionBlock execution
+
+    fun runEffect1(completionBlock: () -> Unit) {
+        completionBlock.invoke()
+    }
+
+    @Test
+    fun `should execute completionBock when such block is provided`() {
```
kleewho, line 72: The fact that the: `should execute completionBock when such block is provided` and `should cancel remoteAction when cancel effect` are in every test files suggest that there's something common across all implementations that maybe it's actually worth separating

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffectTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    private val channels = listOf("channel1")
+    private val channelGroups = listOf("channelGroup1")
+    private val reason = PubNubException("Unknown error")
+    private val attempts = 1
+    private val handshakeProvider = mockk<HandshakeProvider>()
+    private val eventDeliver = SubscribeManagedEffectFactoryTest.TestEventDeliver()
+    private val handshakeReconnectInvocation =
+        SubscribeEffectInvocation.HandshakeReconnect(channels, channelGroups, attempts, reason)
+    private val executorService = Executors.newSingleThreadScheduledExecutor()
+    private val policy = mockk<RetryPolicy>()
```
kleewho, line 25: Instead of setting the mocks everywhere for it why not use the real code and if necessary use other already written policy?

```suggestion
    private val policy = LinearPolicy(duration = Duration.ofMillis(10))
```

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffectTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    fun `should deliver HandshakeReconnectGiveUp event when delay is null`() {
+        // given
+        every { policy.nextDelay(any()) } returns null
+        val handshakeReconnectEffect = HandshakeReconnectEffect(
+            handshakeProvider,
+            eventDeliver,
+            policy,
```
kleewho, line 80: You could use here something already written:

```suggestion
            NoRetriesPolicy,
```

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        // given
+        every { handshakeProvider.handshake(any(), any()) } returns successfulRemoteAction(subscriptionCursor)
+        val handshakeEffect = HandshakeEffect(handshakeProvider, eventDeliver, handshakeInvocation)
+        val completionBlockMock: () -> Unit = mockk()
+        every { completionBlockMock.invoke() } returns Unit
```
kleewho, line 77: If you insist on using `mocks` and `Thread.sleep` you could always save one line here:

```suggestion
        val completionBlockMock = spyk({})
```

But actually better, with the same amount of lines would be:

```
    @Test
    fun `look ma, less mocks`() {
        // given
        every { handshakeProvider.handshake(any(), any()) } returns successfulRemoteAction(subscriptionCursor)
        val handshakeEffect = HandshakeEffect(handshakeProvider, eventDeliver, handshakeInvocation)
        val countDownLatch = CountDownLatch(1)

        // when
        handshakeEffect.runEffect { countDownLatch.countDown() }

        // then
        assertTrue(countDownLatch.await(50, java.util.concurrent.TimeUnit.MILLISECONDS))
    }
```

or if you decide to limit the dependencies of the HandshakeEffect to those that are actually needed then it's even simpler:

```
    @Test
    fun `look ma, no mocks at all`() {
        // given
        val handshakeEffect = HandshakeEffect2(successfulRemoteAction(subscriptionCursor), eventDeliver)
        val countDownLatch = CountDownLatch(1)

        // when
        handshakeEffect.runEffect { countDownLatch.countDown() }

        // then
        assertTrue(countDownLatch.await(50, java.util.concurrent.TimeUnit.MILLISECONDS))
    }
```

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveMessagesEffect.kt
```diff
@@ -0,0 +1,50 @@
...
+        receiveMessagesRemoteAction.silentCancel()
+    }
+
+    private fun getReceiveMessagesRemoteAction(): RemoteAction<ReceiveMessagesResult> =
+        receiveMessagesProvider.receiveMessages(
+            receiveMessagesInvocation.channels,
+            receiveMessagesInvocation.channelGroups,
+            receiveMessagesInvocation.subscriptionCursor
+        )
+}
```
kleewho, line 50: You could change this class so it depends only fewer things. It will make the tests of it simpler, and the implementation is shorter as well. Also it fixes the bug mentioned by me already in other comment.

```suggestion
class ReceiveMessagesEffect(
    private val remoteAction: RemoteAction<ReceiveMessagesResult>,
    private val eventDeliver: EventDeliver
) : ManagedEffect {
    override fun runEffect(completionBlock: () -> Unit) {
        remoteAction.async { result, status ->
            try {
                if (status.error) {
                    eventDeliver.passEventForHandling(
                        Event.ReceiveFailure(
                            status.exception
                                ?: PubNubException("Unknown error") // todo check it that can happen
                        )
                    )
                } else {
                    eventDeliver.passEventForHandling(
                        Event.ReceiveSuccess(
                            result!!.messages,
                            result.subscriptionCursor
                        )
                    )
                }
            } finally {
                completionBlock()
            }
        }
    }

    override fun cancel() {
        val receiveMessagesRemoteAction = remoteAction
        receiveMessagesRemoteAction.silentCancel()
    }
}

```

kleewho, line 50: You could change this class so it depends on fewer things. It will make the tests of it simpler, and the implementation is shorter as well. Also it fixes the bug mentioned by me already in other comment.

```suggestion
class ReceiveMessagesEffect(
    private val remoteAction: RemoteAction<ReceiveMessagesResult>,
    private val eventDeliver: EventDeliver
) : ManagedEffect {
    override fun runEffect(completionBlock: () -> Unit) {
        remoteAction.async { result, status ->
            try {
                if (status.error) {
                    eventDeliver.passEventForHandling(
                        Event.ReceiveFailure(
                            status.exception
                                ?: PubNubException("Unknown error") // todo check it that can happen
                        )
                    )
                } else {
                    eventDeliver.passEventForHandling(
                        Event.ReceiveSuccess(
                            result!!.messages,
                            result.subscriptionCursor
                        )
                    )
                }
            } finally {
                completionBlock()
            }
        }
    }

    override fun cancel() {
        val receiveMessagesRemoteAction = remoteAction
        receiveMessagesRemoteAction.silentCancel()
    }
}

```

kleewho (changes requested):

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffect.kt
marcin-cebo: Make sense. I will modify it.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
marcin-cebo: ok, I like this approach.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveMessagesEffect.kt
marcin-cebo: ok, I will modify them.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffectTest.kt
marcin-cebo: Yes. I can do this.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/RemoteActionManagedEffect.kt
marcin-cebo: Yes, this code was concise. Removal of this class was consequence of moving Effects to separate classes. As discussed we will try to extract some common code for Effect. Maybe at that time code from this class will be used.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
marcin-cebo: Yes. I can fix that.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
marcin-cebo: As discuss we can handle this is separate PR.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffectTest.kt
marcin-cebo: ok, I will use it.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeProvider.kt
```diff
@@ -4,5 +4,5 @@ import com.pubnub.api.endpoints.remoteaction.RemoteAction
 import com.pubnub.api.subscribe.eventengine.event.SubscriptionCursor
 
 fun interface HandshakeProvider {
-    fun handshake(channels: List<String>, channelGroups: List<String>): RemoteAction<SubscriptionCursor>
+    fun getRemoteActionForHandshake(channels: List<String>, channelGroups: List<String>): RemoteAction<SubscriptionCursor>
```
kleewho, line 7: `handshakeRemoteAction`?

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffect.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @Transient
+    private var scheduled: ScheduledFuture<*>? = null
+
+    @Transient
+    private var cancelled = false
+
+    @Synchronized
+    override fun runEffect() {
+        if (cancelled) { // is it ok to be here or should be after "if (delay == null)"
```
kleewho, line 29: Do we need this comment? Because it's synchronized the cancelled can't change during this call so it's better to bail out as quickly as possible

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveMessagesProvider.kt
```diff
@@ -4,7 +4,7 @@ import com.pubnub.api.endpoints.remoteaction.RemoteAction
 import com.pubnub.api.subscribe.eventengine.event.SubscriptionCursor
 
 fun interface ReceiveMessagesProvider {
-    fun receiveMessages(
+    fun getRemoteActionForReceiveMessages(
```
kleewho, line 7: `receiveMessagesRemoteAction`?

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveReconnectEffect.kt
```diff
@@ -0,0 +1,66 @@
...
+    @Transient
+    private var cancelled = false
+    override fun runEffect() {
+        if (cancelled) {
+            return
+        }
+
+        val delay = policy.nextDelay(attempts)
+        if (delay == null) {
+            eventQueue.add(event = Event.ReceiveReconnectGiveUp(reason!!))
```
kleewho, line 33: I don't think we can allow `!!` to be here. We should handle it differently. This might produce `NPE`

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveReconnectEffect.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val attempts: Int,
+    private val reason: PubNubException?
+) : ManagedEffect {
+
+    @Transient
+    private var scheduled: ScheduledFuture<*>? = null
+
+    @Transient
+    private var cancelled = false
+    override fun runEffect() {
```
kleewho, line 26: `@Synchronized` is missing I think in here and in cancel

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -4,46 +4,78 @@ import com.pubnub.api.PubNubException
...
-        failingManagedEffect.runEffect { latch.countDown() }
+        handshakeEffect.runEffect()
 
         // then
-        assertTrue(latch.await(100, TimeUnit.MILLISECONDS))
+        Awaitility.await()
+            .atMost(Durations.ONE_SECOND)
+            .with()
+            .pollInterval(Duration.ofMillis(20))
+            .until { listOf(Event.HandshakeSuccess(subscriptionCursor)) == eventQueue.events }
```
kleewho, line 37: I would rather see assertions here. They can tell us more information in case of failure

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -4,46 +4,78 @@ import com.pubnub.api.PubNubException
...
+            .atMost(Durations.ONE_SECOND)
+            .with()
+            .pollInterval(Duration.ofMillis(20))
+            .until { listOf(Event.HandshakeFailure(reason)) == eventQueue.events }
+    }
+
+    @Test
+    fun `should cancel remoteAction when cancel effect`() {
+        // given
+        val remoteAction: RemoteAction<SubscriptionCursor> = spyk()
```
kleewho, line 59: I'm not sure this is how you suppose to use `spyk`. I'm actually surprised that it works

kleewho:
```suggestion
        val remoteAction: RemoteAction<SubscriptionCursor> = spyk(successfulRemoteAction(subscriptionCursor))
```

kleewho (changes requested):
The only deal breaker is this `Synchronized` missing

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeProvider.kt
marcin-cebo: Ok, I will change it to getHandshakeRemoteAction

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveMessagesProvider.kt
marcin-cebo: Ok. I will change it to getReceiveMessagesRemoteAction.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeReconnectEffect.kt
marcin-cebo: ok, I will delete it.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveReconnectEffect.kt
marcin-cebo: ok. I will add those.

src/main/kotlin/com/pubnub/api/eventengine/EventQueue.kt
```diff
@@ -2,6 +2,7 @@ package com.pubnub.api.eventengine
 
 import com.pubnub.api.subscribe.eventengine.event.Event
 
-interface EventSink {
-    fun put(event: Event)
+interface EventQueue {
+    fun add(event: Event)
+    fun take(): Event
```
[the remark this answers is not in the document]
marcin-cebo: Ok, I will change it to EventSink.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
marcin-cebo: ok, I will change it.

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
marcin-cebo: As discussed I will leave it as is even if it looks like kind of magic.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/ReceiveReconnectEffect.kt
marcin-cebo: Ok, so we can go with:
Event.ReceiveReconnectGiveUp(reason ?: PubNubException("Unknown error"))
What do you think?

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/HandshakeEffectTest.kt
```diff
@@ -4,46 +4,74 @@ import com.pubnub.api.PubNubException
...
-        failingManagedEffect.runEffect { latch.countDown() }
+        handshakeEffect.runEffect()
 
         // then
-        assertTrue(latch.await(100, TimeUnit.MILLISECONDS))
+        Awaitility.await()
+            .atMost(Durations.ONE_SECOND)
+            .with()
+            .pollInterval(Duration.ofMillis(20))
+            .untilAsserted { listOf(Event.HandshakeSuccess(subscriptionCursor)) == eventSink.events }
```
kleewho, line 37: But `untilAsserted` need an assertion of some sort inside. It can't be just a boolean that is being removed.

kleewho (approved):

PR #122, opened 2023-05-08
EmitMessages Effect implementation

marcin-cebo:


src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import com.pubnub.api.models.consumer.PNStatus
+import com.pubnub.api.models.consumer.pubsub.PNMessageResult
+import com.pubnub.api.models.consumer.pubsub.PNPresenceEventResult
+import com.pubnub.api.models.consumer.pubsub.PNSignalResult
+import com.pubnub.api.models.consumer.pubsub.files.PNFileEventResult
+import com.pubnub.api.models.consumer.pubsub.message_actions.PNMessageActionResult
+import com.pubnub.api.models.consumer.pubsub.objects.PNObjectEventResult
+
+interface EmitMessagesProvider {
```
kleewho, line 11: The name `Provider` is usually associated with something that returns something. None of the methods in it return anything so I don't think it should be named `Provider` at all

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun announce(presence: PNPresenceEventResult) // toDo move to Presence EventEngine
+
+    fun announce(signal: PNSignalResult)
+
+    fun announce(messageAction: PNMessageActionResult)
+
+    fun announce(pnObjectEventResult: PNObjectEventResult)
+
+    fun announce(pnFileEventResult: PNFileEventResult)
```
kleewho, line 24: Another question one might have is do we actually want to have this distinction on method level? Do we need so many methods named the same or maybe just one would be enough? 

```
    fun betterMethodNameThenAnnounceMaybe(event: PNEvent)
```

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/BaseEffectTest.kt
```diff
@@ -5,8 +5,8 @@ import com.pubnub.api.models.consumer.pubsub.BasePubSubResult
 import com.pubnub.api.models.consumer.pubsub.PNEvent
 import com.pubnub.api.models.consumer.pubsub.PNMessageResult
 
-open class BaseEffectTest {
-    protected fun createMessages(): List<PNEvent> {
+abstract class BaseEffectTest {
```
kleewho, line 8: Do we need a class at all? Why not just make the `createPNMessageResultList` just a function?

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesEffectTest.kt
```diff
@@ -0,0 +1,242 @@
...
+        pnObjectEventResult: PNObjectEventResult
+    ): List<PNEvent> {
+        val listOfEvents = mutableListOf<PNEvent>()
+        listOfEvents.add(pnMessageResult)
+        listOfEvents.add(pnFileEventResult)
+        listOfEvents.add(pnSignalResult)
+        listOfEvents.add(pnMessageActionResult)
+        listOfEvents.add(pnObjectEventResult)
+
+        return listOfEvents
```
kleewho, line 156:
```suggestion
        return listOf(
            pnMessageResult,
            pnFileEventResult,
            pnSignalResult,
            pnMessageActionResult,
            pnObjectEventResult
        )
```

kleewho (changes requested):

src/main/kotlin/com/pubnub/api/eventengine/EffectDispatcher.kt
```diff
@@ -24,14 +19,14 @@ class EffectDispatcher<T : EffectInvocation>(
 
             is Managed -> {
                 managedEffects.remove(effectInvocation.id)?.cancel()
-                val managedEffect = managedEffectFactory.create(effectInvocation) ?: return
-                managedEffects[effectInvocation.id] = managedEffect
+                val managedEffect = effectFactory.create(effectInvocation) ?: return
+                managedEffects[effectInvocation.id] = managedEffect as ManagedEffect
```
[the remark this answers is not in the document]
marcin-cebo: My understanding is that each invocation has type and code inside  "Managed -> { } " is executed if Invocation has type "Managed".  SubscribeManagedEffectFactory creates Effects and knows how to map invocation to effect of proper type.

To double check we could enhance unit test    

    @Test
    fun `should return handshake effect when getting Handshake invocation`() {
        // when
        val managedEffect =
            subscribeManagedEffectFactory.create(SubscribeEffectInvocation.Handshake(channels, channelGroups))

        // then
        assertThat(managedEffect, instanceOf(HandshakeEffect::class.java))
        assertThat(managedEffect, instanceOf(ManagedEffect::class.java)). <-- add this line to check  type of effect
    }

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesProvider.kt
marcin-cebo: ok, so maybe EmitMessagesConsumer ?

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/BaseEffectTest.kt
marcin-cebo: I wanted do have function that will be used by many classes.
Is static method better design that having abstract class with method like here?

src/test/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesEffectTest.kt
marcin-cebo: I like this. I will make a change.

src/main/kotlin/com/pubnub/api/subscribe/eventengine/effect/EmitMessagesProvider.kt
marcin-cebo: I though that when putting embedding EE into SDK we could have ListenerManager implemented this interface. It has all those method implemented but doesn't have method announce(event: PNEvent)
What do you think?

PR #129, opened 2023-07-04
Incorporate unsubscribe and unsubscribeAll to EventEngine

marcin-cebo:


src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
```diff
@@ -31,7 +35,8 @@ import java.util.concurrent.Executors
 import java.util.concurrent.ScheduledExecutorService
 
 class Subscribe(
-    private val eventEngineManager: EventEngineManager
+    private val eventEngineManager: EventEngineManager,
+    private val stateManager: SubscriptionState = SubscriptionState()
```
marcin-cebo, line 39: Name `stateManager` should be replaced with `subscriptionState`

src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
```diff
@@ -99,36 +104,47 @@ class Subscribe(
         withPresence: Boolean,
         withTimetoken: Long = 0L,
     ) {
+        throwExceptionIfChannelAndChannelGroupIsMissing(channels, channelGroups)
+        addChannelsToLocalStorage(channels)
+        addChannelGroupsToLocalStorage(channelGroups)
```
kleewho, line 109: This method (just like all other methods here) is not synchronized or do not guard the internal state from being modified concurrently and I would expect to see a lot of `ConcurrentModificationExceptions`. As already mentioned in other comment maybe we should think through where exactly should we handle that

kleewho (changes requested):

src/main/kotlin/com/pubnub/api/managers/SubscriptionState.kt
```diff
@@ -0,0 +1,18 @@
+package com.pubnub.api.managers
+
+import com.pubnub.api.models.SubscriptionItem
+
+class SubscriptionState {
+
+    internal var channels: HashMap<String, SubscriptionItem> = hashMapOf()
```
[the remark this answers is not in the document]
marcin-cebo: ok. I will change it.

src/main/kotlin/com/pubnub/api/managers/SubscriptionState.kt
```diff
@@ -0,0 +1,18 @@
...
+        @Synchronized internal set
```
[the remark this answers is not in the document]
marcin-cebo: ok. I will change it.

src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
marcin-cebo: ok. I will change it.

src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
```diff
@@ -3,11 +3,14 @@
...
 
 import com.pubnub.api.PubNub
+import com.pubnub.api.PubNubError
+import com.pubnub.api.PubNubException
 import com.pubnub.api.eventengine.EffectDispatcher
 import com.pubnub.api.eventengine.EventEngineConf
 import com.pubnub.api.eventengine.Sink
 import com.pubnub.api.managers.EventEngineManager
 import com.pubnub.api.managers.ListenerManager
+import com.pubnub.api.managers.SubscriptionData
```
kleewho, line 13: Can we keep it in one package?

src/test/kotlin/com/pubnub/api/subscribe/SubscribeTest.kt
```diff
@@ -0,0 +1,157 @@
...
+
+        assertEquals(
+            Event.SubscriptionChanged(
+                (channelsInSubscriptionData + channelToSubscribe).toList(),
+                (channelGroupsInSubscriptionData + channelGroupsToSubscribe).toList()
+            ),
+            event.captured
+        )
+        assertTrue(subscriptionData.channels.contains(CHANNEL_01))
+        assertTrue(subscriptionData.channelGroups.contains(CHANNEL_GROUPS_01))
```
kleewho, line 58: What does it actually test here?

kleewho: Maybe it's enough to verify that the we have correct event? I mean, we had to get some of the information from somewhere, so if we're able to create an event with correct content then I think we can omit asserting internals

kleewho (approved):

src/test/kotlin/com/pubnub/api/subscribe/SubscribeTest.kt
```diff
@@ -0,0 +1,157 @@
...
+    @Test
+    internal fun `should remove channels and channelGroups from local storage and pass SubscriptionChange event for handling when at least one channel or channelGroup left in storage`() {
+        val channelsToUnsubscribe: List<String> = listOf(CHANNEL_01)
+        val channelGroupsToUnsubscribe: List<String> = listOf(CHANNEL_GROUPS_01)
+
+        objectUnderTest.unsubscribe(channelsToUnsubscribe, channelGroupsToUnsubscribe)
+
+        verify { eventEngineManager.addEventToQueue(any()) }
+        assertEquals(Event.SubscriptionChanged(listOf(CHANNEL_02), listOf(CHANNEL_GROUPS_02)), event.captured)
+        assertTrue(subscriptionData.channels.size == 1 && subscriptionData.channels.contains(CHANNEL_02))
```
kleewho, line 95: If you really have to verify internals there are better ways to write this assertion. 

```
assertEquals(setOf(CHANNEL_02), subscriptionData.channels)
```

src/test/kotlin/com/pubnub/api/subscribe/SubscribeTest.kt
marcin-cebo: ok. I will modify it.

src/test/kotlin/com/pubnub/api/subscribe/SubscribeTest.kt
marcin-cebo: ok. I will modify it.

PR #135, opened 2023-07-26
Added status PNDisconnectedUnexpectedly

marcin-cebo:
CLEN-1437

### DIFF
--- a/src/main/kotlin/com/pubnub/api/enums/PNStatusCategory.kt
+++ b/src/main/kotlin/com/pubnub/api/enums/PNStatusCategory.kt
@@ -99,4 +99,9 @@ enum class PNStatusCategory {
      * Previously started subscribe loop did fail and at this moment client disconnected from real-time data channels.
      */
     PNConnectionError,
+
+    /**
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/pubnub/api/enums/PNStatusCategory.kt
```diff
@@ -99,4 +99,9 @@ enum class PNStatusCategory {
      * Previously started subscribe loop did fail and at this moment client disconnected from real-time data channels.
      */
     PNConnectionError,
+
+    /**
+     * It was not possible for Event Engine to regain connection to PN server.
+     */
+    PNDisconnectedUnexpectedly,
```
kleewho, line 106: I thought that we will reuse already existing status `PNUnexpectedDisconnect`. Having both might be quite confusing

marcin-cebo: Ok. Makes sense. I will change it.

kleewho (approved):

PR #153, opened 2023-11-07
Can read unencrypted history message when crypto is configured.

marcin-cebo:
fix: Fixed reading unencrypted history message when crypto is configured.


src/main/kotlin/com/pubnub/extension/JsonElement.kt
```diff
@@ -4,30 +4,55 @@ import com.google.gson.JsonElement
...
-            mapper.elementToString(this, "pn_other")
+private const val PN_OTHER = "pn_other"
+
+internal fun JsonElement.processHistoryMessage(
+    cryptoModule: CryptoModule?,
+    mapper: MapperManager
+): Pair<JsonElement, String?> {
+
+    cryptoModule ?: return Pair(this, null)
+    val error: String?
```
wkal-pubnub, line 19: not needed here? `error` variable could be local in scopes: line 27 and line 39

src/main/kotlin/com/pubnub/api/models/consumer/history/PNFetchMessage.kt
```diff
@@ -38,7 +38,8 @@ data class PNFetchMessageItem(
     val meta: JsonElement?,
     val timetoken: Long,
     val actions: Map<String, Map<String, List<Action>>>? = null,
-    val messageType: HistoryMessageType?
+    val messageType: HistoryMessageType?,
+    val error: String? = null
```
wkal-pubnub, line 42: I'm not a big fan of adding an `error` field here, using `PNStatus` would seem like a better option for communicating this, however it wouldn't work in the `.sync()` case I think since there's no PNStatus returned, only `Output` or `Exception`... 

If we don't have other options, can we at least not make this a String, and make it an enum instead, with just one possible value (or null) for now? we can add more in the future if we come up with any other possible errors for this

src/main/kotlin/com/pubnub/api/models/consumer/history/PNHistoryResult.kt
```diff
@@ -28,5 +28,6 @@ class PNHistoryResult internal constructor(
 data class PNHistoryItemResult(
     val entry: JsonElement,
     val timetoken: Long? = null,
-    val meta: JsonElement? = null
+    val meta: JsonElement? = null,
+    val error: String? = null
```
wkal-pubnub, line 32: ditto (on the other `error` field)

wkal-pubnub (changes requested):

src/main/kotlin/com/pubnub/api/models/consumer/history/PNFetchMessage.kt
marcin-cebo: I can create entry in PubNubError enum.

src/main/kotlin/com/pubnub/api/models/consumer/history/PNHistoryResult.kt
marcin-cebo: I can create entry in PubNubError enum.

src/main/kotlin/com/pubnub/api/models/consumer/history/PNFetchMessage.kt
```diff
@@ -38,7 +39,8 @@ data class PNFetchMessageItem(
     val meta: JsonElement?,
     val timetoken: Long,
     val actions: Map<String, Map<String, List<Action>>>? = null,
-    val messageType: HistoryMessageType?
+    val messageType: HistoryMessageType?,
+    val error: PubNubError? = null
```
wkal-pubnub, line 43: add documentation (and in the other class) that explains that this can take a value of `CRYPTO_IS_CONFIGURED_BUT_MESSAGE_IS_NOT_ENCRYPTED` (and in what situation) or null

wkal-pubnub (approved):

wkal-pubnub (approved):

marcin-cebo:
@pubnub-release-bot release kotlin as v7.7.2

marcin-cebo:
@pubnub-release-bot release kotlin as v7.7.2

parfeon:
@pubnub-release-bot release kotlin as v7.7.2

PR #155, opened 2023-11-16
Handle unencrypted message in subscribe with crypto

wkal-pubnub:


src/main/kotlin/com/pubnub/api/workers/SubscribeMessageProcessor.kt
```diff
@@ -201,34 +204,56 @@ internal class SubscribeMessageProcessor(
...
+        val inputText = if (pubnub.mapper.isJsonObject(input)) {
+            // property pn_other is used when we want to send encrypted Push Notification, not whole JSON object is encrypted but only value of pn_other property
+            if (pubnub.mapper.hasField(input, PN_OTHER)) {
+                // JSON with pn_other property indicates that this is encrypted Push Notification
+                pubnub.mapper.elementToString(input, PN_OTHER)
+            } else {
+                // plain JSON object indicates that this is not encrypted message
+                return input to logAndReturnDecryptionError()
+            }
+        } else if (input.isJsonPrimitive && input.asJsonPrimitive?.isString == true) {
```
marcin-cebo, line 229: This condition doesn't exist in JsonElement.processHistoryMessage.
What do you think about adding it there ? or maybe we could use JsonElement.processHistoryMessage method here to reduce code duplication?

src/main/kotlin/com/pubnub/api/workers/SubscribeMessageProcessor.kt
```diff
@@ -201,34 +204,56 @@ internal class SubscribeMessageProcessor(
...
+            // String may represent not encrypted string or encrypted data. We will check this when decrypting.
             pubnub.mapper.elementToString(input)
```
marcin-cebo, line 231: If we decide not to use JsonElement.processHistoryMessage here. I would advise to write unit tests for this method similar to those from JsonElement.processHistoryMessage plus extra test to handles:
"else if (input.isJsonPrimitive && input.asJsonPrimitive?.isString == true)"

src/main/kotlin/com/pubnub/api/workers/SubscribeMessageProcessor.kt
wkal-pubnub: you're right, I thought maybe these functions differ somehow, but they are actually exactly the same (or should be)
I'll reuse it

src/main/kotlin/com/pubnub/api/workers/SubscribeMessageProcessor.kt
wkal-pubnub: this case is tested already both in JsonElement test and SubscribeMessageProcessorTest (all cases - including strings, arrays, objects, pn_other)

src/test/kotlin/com/pubnub/extension/JsonElementTest.kt
```diff
@@ -1,67 +1,76 @@
...
         fun cryptoModuleConfiguration(): List<Arguments> {
             val cipherKey = "enigma"
             return listOf(
                 Arguments.of(
-                    "Hello world.",
+                    JsonPrimitive("Hello world."),
                     false,
                     "X/wLZqR/I+4eU7WwtKycTwy9+L9e7m2BhklK9ZQWtwI=",
                     CryptoModule.createLegacyCryptoModule(cipherKey = cipherKey, randomIv = true),
                     generatePayload()
```
marcin-cebo, line 30: I guess we can remove 'generatePayload()' from all Arguments.

marcin-cebo (approved):

src/test/kotlin/com/pubnub/extension/JsonElementTest.kt
wkal-pubnub: Done

PR #157, opened 2023-11-22
Make channels and channelGroups in Subscribe EE immutable.

marcin-cebo:


src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
```diff
@@ -94,8 +94,8 @@ internal class Subscribe(
         throwExceptionIfChannelAndChannelGroupIsMissing(channels, channelGroups)
         addChannelsToSubscriptionData(channels, withPresence)
         addChannelGroupsToSubscriptionData(channelGroups, withPresence)
-        val channelsInLocalStorage = subscriptionData.channels
-        val channelGroupsInLocalStorage = subscriptionData.channelGroups
+        val channelsInLocalStorage = subscriptionData.channels.toSet()
```
wkal-pubnub, line 97: line 129 probably needs the same treatment

wkal-pubnub (changes requested):

wkal-pubnub (approved):

PR #161, opened 2023-11-27
Add error field to file result and set on decrypt error

wkal-pubnub:
feat: Add `error` field to `PNFileEventResult` and set it in case of decryption failure


wkal-pubnub:
@pubnub-release-bot release kotlin as v7.7.4

PR #162, opened 2023-11-29
Make channels and channelGroups immutable in Presence EE.

marcin-cebo:


src/main/kotlin/com/pubnub/api/presence/Presence.kt
```diff
@@ -132,9 +132,9 @@ internal class EnabledPresence(
         connected: Boolean
     ) {
         if (connected) {
-            joined(channels, channelGroups)
+            joined(channels.toSet(), channelGroups.toSet())
```
wkal-pubnub, line 135: you're copying the sets twice now (here, and inside `joined`/`left`)

How about instead of making this a responsibility of the caller, fix the `PresenceEvent.Joined` and `PresenceEvent.Left` classes to make a copy of the sets internally?

that way we won't make this error again if writing any new code that calls the constructor

(it's probably fine to make them regular classes and not data classes, you can always generate an equals+hashcode if you need it, but it doesn't look like we're depending on it anywhere)

wkal-pubnub (changes requested):

src/main/kotlin/com/pubnub/api/presence/Presence.kt
marcin-cebo: Good point. I like this suggestion. I will modify the code.

PR #164, opened 2023-12-01
Add GSON serialization behavior from Java SDK

wkal-pubnub:
fix: allow serialization of anonymous collection classes like in the PN Java SDK


marcin-cebo (approved):
LGTM

PR #167, opened 2023-12-08
Send state with Subscribe calls with TT=0

wkal-pubnub:


src/main/kotlin/com/pubnub/api/PNConfiguration.kt
```diff
@@ -251,6 +251,20 @@ open class PNConfiguration(
...
+     *
+     * Defaults to `true`.
+     *
+     * Please note that `sendStateWithSubscribe` doesn't apply to state that was set on channel groups.
+     * It is recommended to disable this option if you set state for channel groups using [PubNub.setPresenceState],
+     * otherwise that state may be overwritten on the next subscribe reconnection with individual channel states.
+     */
+    var sendStateWithSubscribe = true
+
     /**
```
marcin-cebo, line 268: This is side comment.

As we discussed. If possible I would be beneficial to add tests that demonstrate purpose or sending state with subscribe tt=0 and hearbeat.

marcin-cebo (approved):

PR #169, opened 2023-12-15
Rename and combine configuration options to maintainPresenceState

wkal-pubnub:


src/main/kotlin/com/pubnub/api/PNConfiguration.kt
```diff
@@ -239,31 +239,18 @@ open class PNConfiguration(
...
-     * Applies only if [enableEventEngine] is true.
-     *
-     * Defaults to `true`.
-     *
-     * Please note that `sendStateWithSubscribe` doesn't apply to state that was set on channel groups.
-     * It is recommended to disable this option if you set state for channel groups using [PubNub.setPresenceState],
-     * otherwise that state may be overwritten on the next subscribe reconnection with individual channel states.
-     */
-    var sendStateWithSubscribe = true
+    var maintainPresenceState = true
```
marcin-cebo, line 253: Is this somewhere documented that default is true ?

Going with default (maintainPresenceState = true) client may experience strange behaviour of SDK when setting state for channel groups. Are we ok with this?

src/main/kotlin/com/pubnub/api/PubNub.kt
```diff
@@ -738,10 +738,10 @@ class PubNub internal constructor(
      *
      * State information is supplied as a JSON object of key/value pairs.
      *
-     * If [PNConfiguration.sendStateWithHeartbeat] is `true`, the state for channels will be saved in the PubNub
-     * client and resent with every heartbeat. In that case, it's not recommended to mix setting state
+     * If [PNConfiguration.maintainPresenceState] is `true`, the state for channels will be saved in the PubNub
+     * client and resent with every heartbeat and resubscribe. In that case, it's not recommended to mix setting state
      * through channels *and* channel groups, as state set through the channel group will be overwritten
-     * after the next heartbeat.
+     * after the next heartbeat or resubscribe.
```
marcin-cebo, line 744: Is "resubscribe" the same as "subscribe reconnection"?

I am asking because in PNConfiguration there is:  
 "for the current [userId] with every automatic heartbeat (if [heartbeatInterval] is greater than 0) and subscribe reconnection."

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
wkal-pubnub: As far as I remember we already had this behavior in Kotlin in non-EE version, so that's why it's on by default

src/main/kotlin/com/pubnub/api/PubNub.kt
wkal-pubnub: hmm what's the best way to describe the underlying subscribe calls (the automatic ones that happen every 300 seconds, not necessarily in response to any user facing api like reconnect() or subscribe())

wkal-pubnub: maybe just "subscribe connection" ? without going into details?

src/main/kotlin/com/pubnub/api/PNConfiguration.kt
marcin-cebo: I see.

src/main/kotlin/com/pubnub/api/PubNub.kt
marcin-cebo: LGTM.

marcin-cebo (approved):
LGTM

PR #175, opened 2024-01-19
Remove saved state internally even with Presence EE disabled

wkal-pubnub:


src/main/kotlin/com/pubnub/api/subscribe/Subscribe.kt
```diff
@@ -134,6 +135,8 @@ internal class Subscribe(
         removeChannelsFromSubscriptionData(channels)
         removeChannelGroupsFromSubscriptionData(channelGroups)
 
+        presenceData.channelStates.keys.removeAll(channels)
+
```
marcin-cebo, line 139: I am wondering. Currently we have two classes SubscriptionData and PresenceData 
Do we need both of them? Maybe we could merge them ? 

e.g.  we could remove PresenceData and modify SubscriptionData
internal class SubscriptionData {
    internal val channelsWithState: ConcurrentHashMap<String, Any> = ConcurrentHashMap()
    internal val channelGroups: MutableSet<String> = mutableSetOf()
}
?

wkal-pubnub: I considered it but I think I'd rather not couple these two features. For example, if we wanted to move this again, or for some reason add this to the non-EE subscribe loop (not likely but still...), or I don't know... make it work different independent of channel subscriptions... it's easier to do if we're not coupling it with subscription data. OK to keep it separate?

marcin-cebo: Yes.

marcin-cebo (approved):

PR #176, opened 2024-02-02
swap hashmap for concurrentMap and use threadsafe operations

makosblade:
This PR:
- replaces HashMap with a ConcurrentMap in TelemetryManager
- uses threadsafe composite operations in place of previous multistep operations


ryanmanns1-pubnub (approved):

wkal-pubnub:
Thanks for the contribution, but we've learned since that `TelemetryManager` and the whole telemetry feature will be removed in the next version of the SDK. Given this information, there's no need to merge this PR.

PR #177, opened 2024-02-06
Fixed transitions so that SubscribeEvenEngine can handle multiple…

marcin-cebo:
fix: Fixed transitions so that SubscribeEvenEngine can handle multiple subscribe calls executed one by one.

fix: Add missing Leave events with EE enabled and heartbeat = 0


src/integrationTest/kotlin/com/pubnub/api/integration/SubscribeIntegrationTests.kt
```diff
@@ -115,6 +117,55 @@ class SubscribeIntegrationTests : BaseIntegrationTest() {
...
+
+            override fun message(pubnub: PubNub, pnMessageResult: PNMessageResult) {
+                println("-=message ${pnMessageResult.message}") // <-- here I am able to get msgFromUser1ToChannel01 and msgFromUser2ToChannel01
+                assertTrue(listOf(expectedMessage01, expectedMessage02, expectedMessage03).contains(pnMessageResult.message.asString))
+                countDownLatch.countDown()
+            }
+        })
+
+        pubnub1.subscribe(channels = listOf(channel01), withTimetoken = timeToken10minAgo)
+        pubnub1.publish(channel01, expectedMessage03).sync()
```
wkal-pubnub, line 158: is this publish necessary here to trigger the error condition (before the fix?) 
I'm worried that it might introduce an unnecessary delay between subscribes, if it's not needed consider removing

wkal-pubnub (approved):

src/integrationTest/kotlin/com/pubnub/api/integration/SubscribeIntegrationTests.kt
marcin-cebo: I tested before and without the fixed code  a presence of this additional publish doesn't effect result.
Additionally I would like to test that subscribe call that follows this publish doesn't cause the message from this publish being not deliver to listener.

marcin-cebo:
@pubnub-release-bot release kotlin as v7.8.1

marcin-cebo:
@pubnub-release-bot release kotlin as v7.8.1

PR #178, opened 2024-02-06
Add missing Leave events with EE enabled and heartbeat = 0

wkal-pubnub:


src/test/kotlin/com/pubnub/api/presence/PresenceTest.kt
```diff
@@ -94,20 +97,70 @@ internal class PresenceTest {
         assertThat(presence, Matchers.isA(PresenceNoOp::class.java))
     }
 
+    @Test
+    fun `Leave events not created when suppressLeaveEvents is true and heartbeat interval is 0`() {
+        // given
+        val presence = Presence.create(
+            listenerManager = listenerManager, heartbeatInterval = 0.seconds, suppressLeaveEvents = true,
+            leaveProvider = object : LeaveProvider {
```
marcin-cebo, line 105: We could use mock of LeaveProvider like this:

```
    @Test
    fun `Leave events not created when suppressLeaveEvents is true and heartbeat interval is 0`() {
        val leaveProviderMock: LeaveProvider = mockk()
        // given
        val presence = Presence.create(
            listenerManager = listenerManager, heartbeatInterval = 0.seconds, suppressLeaveEvents = true,
            leaveProvider = leaveProviderMock
        )

        // when
        presence.joined(setOf("abc"))
        presence.leftAll()

        // then
        verify(exactly = 0) { leaveProviderMock.getLeaveRemoteAction(any(), any()) }
    }
```

What do you think?

src/test/kotlin/com/pubnub/api/presence/PresenceTest.kt
```diff
@@ -94,20 +97,70 @@ internal class PresenceTest {
...
+        // no exception
+    }
+
+    @Test
+    fun `Leave events created when suppressLeaveEvents is false and heartbeat interval is 0`() {
+        // given
+        val success = AtomicBoolean(false)
+        val presence = Presence.create(
+            listenerManager = listenerManager, heartbeatInterval = 0.seconds, suppressLeaveEvents = false,
+            leaveProvider = object : LeaveProvider {
```
marcin-cebo, line 129: We could use mock here. What do you think?

src/main/kotlin/com/pubnub/api/presence/Presence.kt
```diff
@@ -33,7 +33,7 @@ internal interface Presence {
             executorService: ScheduledExecutorService
         ): Presence {
             if (heartbeatInterval <= Duration.ZERO || !enableEventEngine) {
-                return PresenceNoOp()
+                return PresenceNoOp(if (!suppressLeaveEvents) leaveProvider else null)
```
marcin-cebo, line 36: I am thinking to pass to PresenceNoOp both suppressLeaveEvents and leaveProvider and inside PresenceNoOp have logic that is responsible for sending leave event. 
That way PresenceNoOp is responsible for sending leave event if needed instead of external code.
What do you think?

wkal-pubnub: done

src/test/kotlin/com/pubnub/api/presence/PresenceTest.kt
wkal-pubnub: done

src/test/kotlin/com/pubnub/api/presence/PresenceTest.kt
wkal-pubnub: done

marcin-cebo (approved):

PR #182, opened 2024-02-12
Added possibility to define specific listener as property.

marcin-cebo:


src/main/kotlin/com/pubnub/internal/v2/subscription/SubscriptionSetImpl.kt
```diff
@@ -22,6 +28,43 @@ internal class SubscriptionSetImpl(
     private val pubnub: PubNub,
     initialSubscriptions: Set<SubscriptionImpl> = emptySet(),
 ) : SubscriptionSet(), EventEmitter {
+
+    override var onMessage: ((PubNub, PNMessageResult) -> Unit)? = null
```
wkal-pubnub, line 32: all these could be delegated properties:
```
    override var onMessage: ((PubNub, PNMessageResult) -> Unit)? by eventEmitter::onMessage
    override var onPresence: ((PubNub, PNPresenceEventResult) -> Unit)? by eventEmitter::onPresence
    override var onSignal: ((PubNub, PNSignalResult) -> Unit)? by eventEmitter::onSignal
    override var onMessageAction: ((PubNub, PNMessageActionResult) -> Unit)? by eventEmitter::onMessageAction
    override var onObjects: ((PubNub, PNObjectEventResult) -> Unit)? by eventEmitter::onObjects
    override var onFile: ((PubNub, PNFileEventResult) -> Unit)? by eventEmitter::onFile
```

src/main/kotlin/com/pubnub/internal/v2/callbacks/EventEmitterImpl.kt
```diff
@@ -22,6 +22,122 @@ internal class EventEmitterImpl(
 
     private val listeners = CopyOnWriteArraySet<EventListener>()
 
+    @Volatile
```
wkal-pubnub, line 25: you could greatly simplify the code and lose all the synchronization with this trick:
```
    override var onMessage: ((PubNub, PNMessageResult) -> Unit)? = null
    override var onPresence: ((PubNub, PNPresenceEventResult) -> Unit)? = null
    override var onSignal: ((PubNub, PNSignalResult) -> Unit)? = null
    override var onMessageAction: ((PubNub, PNMessageActionResult) -> Unit)? = null
    override var onObjects: ((PubNub, PNObjectEventResult) -> Unit)? = null
    override var onFile: ((PubNub, PNFileEventResult) -> Unit)? = null

    private val pluggableListener = object : EventListener {
        override fun message(pubnub: PubNub, result: PNMessageResult) {
            onMessage?.invoke(pubnub, result)
        }
        override fun presence(pubnub: PubNub, result: PNPresenceEventResult) {
            onPresence?.invoke(pubnub, result)
        }
        override fun signal(pubnub: PubNub, result: PNSignalResult) {
            onSignal?.invoke(pubnub, result)
        }
        override fun messageAction(pubnub: PubNub, result: PNMessageActionResult) {
            onMessageAction?.invoke(pubnub, result)
        }
        override fun objects(pubnub: PubNub, result: PNObjectEventResult) {
            onObjects?.invoke(pubnub, result)
        }
        override fun file(pubnub: PubNub, result: PNFileEventResult) {
            onFile?.invoke(pubnub, result)
        }
    }.apply {
        addListener(this)
    }
```

I haven't checked if this works though

src/main/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
```diff
@@ -22,6 +28,41 @@ internal class SubscriptionImpl(
     options: SubscriptionOptions? = null,
 ) : Subscription(), EventEmitter {
 
+    override var onMessage: ((PubNub, PNMessageResult) -> Unit)? = null
```
wkal-pubnub, line 31: same comment as SubSetImpl: use delegated properties

src/main/kotlin/com/pubnub/api/PubNub.kt
```diff
@@ -161,6 +170,7 @@ class PubNub internal constructor(
     internal val tokenManager: TokenManager = TokenManager()
     private val tokenParser: TokenParser = TokenParser()
     internal val listenerManager = ListenerManager(this)
+    private val eventEmitter = EventEmitterImpl(AnnouncementCallback.Phase.SUBSCRIPTION, ::accepts) // todo I tried to used it but doesn't work yet
```
wkal-pubnub, line 173: maybe you forgot to `listenerManager.addAnnouncementCallback(eventEmitter)` ?

src/main/kotlin/com/pubnub/api/PubNub.kt
```diff
@@ -174,6 +184,8 @@ class PubNub internal constructor(
         executorService
     )
 
+    private fun accepts(envelope: AnnouncementEnvelope<*>) = true
```
wkal-pubnub, line 187: you can move this to line 173 simply as the second argument: `{ true }`

wkal-pubnub (changes requested):

src/main/kotlin/com/pubnub/internal/v2/subscription/SubscriptionSetImpl.kt
marcin-cebo: Nice, so concise. I will change it.

src/main/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
marcin-cebo: ok, I will change it.

src/main/kotlin/com/pubnub/api/PubNub.kt
marcin-cebo: ok, I will move it.

src/main/kotlin/com/pubnub/api/PubNub.kt
marcin-cebo: Thanks. That was it.

src/main/kotlin/com/pubnub/internal/v2/callbacks/EventEmitterImpl.kt
marcin-cebo: Please, have a look at changes. They are kind of in line with what you proposed but not exactly.

wkal-pubnub: I don't really like the proposed solution for the following reasons:
- it still needs synchronization
- it's much longer and has more complex code
- it adds casting

Are there any advantages to the solution you proposed vs the single listener approach in my comment above, or reasons why my solution wouldn't work?

wkal-pubnub (approved):

src/main/kotlin/com/pubnub/api/v2/callbacks/EventEmitter.kt
```diff
@@ -10,12 +9,144 @@ import com.pubnub.api.models.consumer.pubsub.message_actions.PNMessageActionResu
...
+     */
+    var onMessage: ((PNMessageResult) -> Unit)?
+
+    /**
+     * A nullable property designed to set a function or lambda expression for handling incoming presence events.
+     * This function is called whenever a new presence event occurs, offering an efficient method for tracking presence updates.
+     *
+     * Note that this property allows for the assignment of a singular behavior at a time, as any new assignment will override the previous one.
+     * For scenarios requiring multiple behaviors in response to presence events, it is advisable
+     * to utilize `com.pubnub.api.PubNub.addListener(listener: EventListener)`.
```
wkal-pubnub, line 42: everywhere you have this, you could do this instead:
```
to utilize [EventEmitter.addListener].
```

because we don't want to point people from using e.g. Subscription.onMessage to PubNub.addListener (which can be implied by your version)

PR #188, opened 2024-02-21
Send Leave when explicitly calling presence(connected=false) with EE

wkal-pubnub:
feat: A new version of subscription and presence handling is enabled by default (`enableEventEngine` flag is set to `true`). Please consult the documentation for new PNStatus values that are emitted for subscriptions, as code changes might be required to support this change.

feat: Added support for scoped event listeners using new entity objects: Channels, ChannelGroups, ChannelMetadata and UserMetadata.


marcin-cebo (approved):

wkal-pubnub:
@pubnub-release-bot release kotlin as v8.0.0

PR #189, opened 2024-03-12
Added convenient method to handle events.

marcin-cebo:


pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
```diff
@@ -36,6 +50,10 @@ class SubscriptionImpl(
         return pubnub.subscriptionSetOf(setOf(this, subscription))
     }
 
+    override fun subscribe() {
+        super.subscribe(SubscriptionCursor(0))
```
wkal-pubnub, line 54: why super? just `subscribe(SubscriptionCursor(0))`
if we ever override it in this class, the no-argument function should use the override, not the superclass implementation


this applies to other places in this PR where `super.XXX` is called

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
```diff
@@ -51,4 +69,85 @@ class SubscriptionImpl(
             }
         }
     }
+
+    override fun setOnMessage(onMessageHandler: OnMessageHandler?) {
+        setOnMessageHandler(onMessageHandler)
```
wkal-pubnub, line 74: why add a layer of indirection, why not put `emitterHelper.onMessage =` here?

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/EmitterHelper.kt
```diff
@@ -0,0 +1,90 @@
...
+                        result: PNFileEventResult,
+                    ) {
+                        onFile?.invoke(result)
+                    }
+                },
+            ),
+        )
+    }
+
+    var onMessage: ((PNMessageResult) -> Unit)? = null
```
wkal-pubnub, line 82: these could use the types you created, for example `var onMessage: OnMessageHandler? = null`
no need to use Kotlin functional types that return Unit
that would also let you simplify code that sets these handlers in the other classes (see other comment)

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
```diff
@@ -51,4 +69,85 @@ class SubscriptionImpl(
...
+    override fun setOnMembership(onMembershipHandler: OnMembershipHandler?) {
+        setOnMembershipHandler(onMembershipHandler)
+    }
+
+    override fun setOnFile(onFileHandler: OnFileHandler?) {
+        setOnFileHandler(onFileHandler)
+    }
+
+    private fun setOnMessageHandler(onMessageHandler: OnMessageHandler?) {
+        emitterHelper.onMessage = onMessageHandler?.let { handler ->
```
wkal-pubnub, line 106: if you change the type in emitterHelper to OnMessageHandler, this could be just
`emitterHelper.onMessage = onMessageHandler`

(see comment in EmitterHelper)

wkal-pubnub (changes requested):

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
marcin-cebo: Good point. I will remove this additional layer.

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/EmitterHelper.kt
marcin-cebo: Good point. I will change it.

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
marcin-cebo: Got it.

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
marcin-cebo: For me:
super.subscribe(SubscriptionCursor(0))
is safer and straightforward. It protects from accidental recursion. We can always change it to 
subscribe(SubscriptionCursor(0)) in the future when there is a need to override subscribe(cursor: SubscriptionCursor) in this class.

What do you think?

wkal-pubnub: the problem is that the compiler won't tell someone in the future that this needs to be changed...
and I don't think we're gaining anything now by calling `super` other than potential problems later.
can you explain why it's safer?

even if you look at the code generated by kotlin @JvmOverloads (if we could use it here), it would always generate `this.subscribe(...)`, because that's just the desired behaviour that we're trying to achieve here.

marcin-cebo: I think I see your point.

Problematic situation might be:
```
    override fun subscribe() {
        subscribe(SubscriptionCursor(0))
    }

    override fun subscribe(cursor: SubscriptionCursor) {
        subscribe(cursor)
    }

```
For me both approaches(using super or not) have pros and cons. 
I can go with your proposal.

marcin-cebo: Other thing that I don't like is when there is:
`subscribe(SubscriptionCursor(0))`
and you click on subscribe method it shows two options.

PR #190, opened 2024-03-13
Moved EventEmitter to constructor.

marcin-cebo:
I have just moved event emitter. For now I don't have simple solution how to move emitterHelper to constructor.

### DIFF
--- a/pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/v2/subscription/BaseSubscriptionImpl.kt
+++ b/pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/v2/subscription/BaseSubscriptionImpl.kt
@@ -20,6 +20,9 @@ abstract class BaseSubscriptionImpl<T : BaseEventListener>(
     channels: Set<ChannelName> = emptySet(),
     channelGroups: Set<ChannelGroupName> = emptySet(),
     options: SubscriptionOptions? = null,
+    eventEmitterFactory: (BaseSubscriptionImpl<T>) -> EventEmitterImpl = { baseSubscriptionImpl ->
+        EventEmitterImpl(AnnouncementCallback.Phase.SUBSCRIPTION, baseSubscriptionImpl::accepts)
+    },
 ) : BaseSubscription<T> {
     @Volatile
     var isActive = false
@@ -46,9 +49,10 @@ abstract class BaseSubscriptionImpl<T : BaseEventListener>(
      */
     private var lastTimetoken: Long = 0L
 
-    protected val eventEmitter = EventEmitterImpl(AnnouncementCallback.Phase.SUBSCRIPTION, ::accepts)
+    protected val eventEmitter = eventEmitterFactory(this)
 
-    private fun accepts(envelope: AnnouncementEnvelope<out PNEvent>): Boolean {
+    // todo changing to public probably is not a big problem since BaseSubscriptionImpl and derived classes are not expose to the user
+    fun accepts(envelope: AnnouncementEnvelope<out PNEvent>): Boolean {
         val event = envelope.event
         val accepted = isActive && filters.all { filter -> filter.predicate(event) } && checkAndUpdateTimetoken(event)
         if (accepted) {
--- a/pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/v2/subscription/BaseSubscriptionImpl.kt
+++ b/pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/v2/subscription/BaseSubscriptionImpl.kt
@@ -51,7 +51,6 @@ abstract class BaseSubscriptionImpl<T : BaseEventListener>(
 
     protected val eventEmitter = eventEmitterFactory(this)
 
-    // todo changing to public probably is not a big problem since BaseSubscriptionImpl and derived classes are not expose to the user
     fun accepts(envelope: AnnouncementEnvelope<out PNEvent>): Boolean {
         val event = envelope.event
         val accepted = isActive && filters.all { filter -> filter.predicate(event) } && checkAndUpdateTimetoken(event)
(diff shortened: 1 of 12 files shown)


pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/v2/subscription/BaseSubscriptionImpl.kt
```diff
@@ -46,9 +49,10 @@ abstract class BaseSubscriptionImpl<T : BaseEventListener>(
      */
     private var lastTimetoken: Long = 0L
 
-    protected val eventEmitter = EventEmitterImpl(AnnouncementCallback.Phase.SUBSCRIPTION, ::accepts)
+    protected val eventEmitter = eventEmitterFactory(this)
 
-    private fun accepts(envelope: AnnouncementEnvelope<out PNEvent>): Boolean {
+    // todo changing to public probably is not a big problem since BaseSubscriptionImpl and derived classes are not expose to the user
```
wkal-pubnub, line 54: yeah that's fine, you can remove the todo

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImplTest.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val listener: EventListenerCore = mockk()
+
+    @BeforeEach
+    internal fun setUp() {
+        every { pubnub.pubNubCore } returns pubnubCore
+        objectUnderTest = TestableSubscriptionImpl(pubnub, channels, channelGroups, options) { eventEmitter }
+    }
+
+    @Test
+    fun `addListener should call addListener on eventEmitter`() {
```
wkal-pubnub, line 37: this method isn't testing SubscriptionImpl, it's testing `BaseSubscriptionImpl.addListener(EventListenerCore)`


if you wanted to test `SubscriptionImpl.addListener(EventListener)`, you could check if it calls `addListener(EventListenerCore)` and passes in a `DelegatingEventListener` that is equals to `DelegatingEventListener(listener)` where listener is an `EventListener`

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImplTest.kt
```diff
@@ -0,0 +1,77 @@
...
+    private class TestableSubscriptionImpl(
+        pubnub: PubNubImpl,
+        channels: Set<ChannelName>,
+        channelGroups: Set<ChannelGroupName>,
+        options: SubscriptionOptions,
+        eventEmitterFactory: (BaseSubscriptionImpl<EventListener>) -> EventEmitterImpl,
+    ) : BaseSubscriptionImpl<EventListener>(pubnub.pubNubCore, channels, channelGroups, options, eventEmitterFactory) {
+        fun testAddListener(listener: EventListenerCore) = addListener(listener)
+
+        fun testRemoveListener(listener: Listener) = removeListener(listener)
```
wkal-pubnub, line 67: why do we need these test* methods?

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
```diff
@@ -26,7 +28,11 @@ class SubscriptionImpl(
     channels: Set<ChannelName>,
     channelGroups: Set<ChannelGroupName>,
     options: SubscriptionOptions,
-) : Subscription, BaseSubscriptionImpl<EventListener>(pubnub.pubNubCore, channels, channelGroups, options) {
+    eventEmitterFactory: (BaseSubscriptionImpl<EventListener>) -> EventEmitterImpl = { baseSubscriptionImpl ->
+        EventEmitterImpl(AnnouncementCallback.Phase.SUBSCRIPTION, baseSubscriptionImpl::accepts)
+    },
+) : Subscription,
+    BaseSubscriptionImpl<EventListener>(pubnub.pubNubCore, channels, channelGroups, options, eventEmitterFactory) {
     // todo maybe we can add it to constructor so we can mock it?
```
wkal-pubnub, line 36: would it be worth doing it in the same PR to check if it's possible now, with the new changes? unless you've checked and you know it's possible without making further changes to the eventEmitter

wkal-pubnub (changes requested):

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImplTest.kt
marcin-cebo: I guess we need only for addListener because it is protected in BaseSubscriptionImpl so you can't do:
```
private val evenListenerCoreSlot = slot<EventListenerCore>()
every { objectUnderTest.addListener(evenListenerCoreSlot.captured) } returns Unit
```
because you will get:
`Cannot access 'addListener': it is protected in 'SubscriptionImpl'`

wkal-pubnub: in that case, should it be tested here? or in BaseSubscriptionImplTest in pubnub-core? 
(see my other comment)

pubnub-gson/src/main/java/com/pubnub/internal/v2/subscription/SubscriptionImpl.kt
wkal-pubnub: I just saw your description "For now I don't have simple solution how to move emitterHelper to constructor"

let me take a quick look

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/subscription/SubscriptionImplTest.kt
marcin-cebo: I found a way to make a test without adding TestableSubscriptionImpl

PR #191, opened 2024-03-15
Added additional constructors for java usages.

marcin-cebo:


pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
```diff
@@ -40,6 +40,11 @@ sealed class RetryConfiguration {
             excludedOperations: List<RetryableEndpointGroup> = emptyList(),
         ) : this(delayInSec.seconds, maxRetryNumber, excludedOperations, false)
 
+        // additional constructors for java
+        constructor() : this(MIN_DELAY, MAX_RETRIES_IN_LINEAR, emptyList())
```
wkal-pubnub, line 44: would `@JvmOverloads` on the secondary constructor work here? I think it will produce the same effect

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
```diff
@@ -65,21 +70,37 @@ sealed class RetryConfiguration {
...
-            minDelayInSec: Int = MIN_DELAY, // min value is 2
-            maxDelayInSec: Int = MAX_DELAY, // max value is 150
-            maxRetryNumber: Int = MAX_RETRIES_IN_EXPONENTIAL, // max value is 6
+            minDelayInSec: Int = MIN_DELAY,
+            maxDelayInSec: Int = MAX_DELAY,
+            maxRetryNumber: Int = MAX_RETRIES_IN_EXPONENTIAL,
             excludedOperations: List<RetryableEndpointGroup> = emptyList(),
         ) : this(minDelayInSec.seconds, maxDelayInSec.seconds, maxRetryNumber, excludedOperations, false)
 
+        // additional constructors for java
```
wkal-pubnub, line 88: same here

wkal-pubnub (changes requested):

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
marcin-cebo: Unfortunately @JvmOverloads doesn't work for me here because it doesn't generate 
`constructor() : this(MIN_DELAY, MAX_RETRIES_IN_LINEAR, emptyList())`
and if I try to add it manually I get:
```
Platform declaration clash: The following declarations have the same JVM signature ( ()V): 
public constructor Linear() defined in com.pubnub.api.retry.RetryConfiguration.Linear
public constructor Linear(delayInSec: Int = ..., maxRetryNumber: Int = ..., excludedOperations: List<RetryableEndpointGroup> = ...) defined in com.pubnub.api.retry.RetryConfiguration.Linear
```

pubnub-gson/src/integrationTest/java/com/pubnub/api/integration/RetryConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void canCreateRetryConfigurationsWithDifferentParameters() {
+        int delayInSec = 2;
+        int minDelay = 3;
+        int maxDelay = 5;
+        int maxRetryNumber = 2;
+        List<RetryableEndpointGroup> excludedOperations = new ArrayList<>();
+
+        RetryConfiguration linearRetryConfigurationWithNoParams = new RetryConfiguration.Linear();
```
marcin-cebo, line 53: I found other strange thing. I am able to do:
RetryConfiguration linearRetryConfigurationWithNoParams = new RetryConfiguration.Linear(1L, 1, excludedOperations, true);
without compilation error and this constructor in defined as internal in different module.

Please, could you check if it behaves the same in you IDE?

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
```diff
@@ -65,21 +70,37 @@ sealed class RetryConfiguration {
      * @property excludedOperations List of [RetryableEndpointGroup] to be excluded from retry.
      */
     class Exponential internal constructor(
-        var minDelayInSec: Duration = MIN_DELAY.seconds, // min value is 2
```
marcin-cebo, line 68: This is also strange to me. I remove this comment ( // min value is 2) because running
` ./gradlew ktlintCheck --rerun-tasks`
or
` ./gradlew clean build`
gives error like:
"A comment in a 'value_parameter_list' is only allowed when placed on a separate line (cannot be auto-corrected) (standard:discouraged-comment-location)
"
but at that time it didn't pointed out that the same issue is 38, 39 ,40 so the build passed successfully.

Then I did some other changes and getting error for 38, 39,40 lines :|

pubnub-gson/src/integrationTest/java/com/pubnub/api/integration/RetryConfigurationIntegrationTest.java
wkal-pubnub: Java doesn't know about Kotlin internal, it treats it as public. I'll see if I can deal with this in another way

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
wkal-pubnub: OK, so I tested it and for some weird reason the IDE can't see the generated constructors, but they're there. We could release the library and users would see it. But let's leave your version for now to not have the error in the IDE. This must be an intelliJ/Kotlin bug, maybe worth reporting

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
wkal-pubnub: either disable the ktlint rule, or move the comment above the line like so:
```
// min value is 2
var minDelayInSec: Duration = MIN_DELAY.seconds, 
```

it's possible that these lines were part of the generated baseline to ignore this error, but after editing the line numbers changed and the baseline file no longer matched

wkal-pubnub (approved):

pubnub-gson/src/integrationTest/java/com/pubnub/api/integration/RetryConfigurationIntegrationTest.java
wkal-pubnub: I have this fixed locally, I'll make the change after we merge your PR
If you're wondering the trick is to make this constructor private, and create a factory method for tests in a companion object that is marked `@JvmSynthetic`. If only Kotlin supported synthetic constructors 🙄

PR #193, opened 2024-03-25
Move retry outside EventEngine.

marcin-cebo:


wkal-pubnub (approved):
This PR LGTM, but overall I think it's not great to tie together retry policy for EE and for one-off actions, especially related to publishing (where it can be desirable to fail fast as opposed to a long-poll loop). I'd like us to think of a way to make it more flexible, for example override retry per-endpoint, or per call (in the future)

PR #194, opened 2024-03-26
Immutable config & builder

wkal-pubnub:


pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/v2/BasePNConfiguration.kt
```diff
@@ -0,0 +1,528 @@
...
+import javax.net.ssl.SSLSocketFactory
+import javax.net.ssl.X509ExtendedTrustManager
+
+/**
+ * A storage for user-provided information which describe further PubNub client behaviour.
+ * Configuration instance contains additional set of properties which
+ * allow performing precise PubNub client configuration.
+ *
+ */
+interface BasePNConfiguration {
```
marcin-cebo, line 24: What is  the value of having this interface and BasePNConfigurationImpl?

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/PNConfigurationImplTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    fun testDefaultTimeoutValues() {
+        val config = PNConfiguration.builder(UserId(BasePubNubImpl.generateUUID()), "demo")
+        assertEquals(300, config.presenceTimeout)
+        assertEquals(0, config.heartbeatInterval)
+    }
+
+    @Test
+    fun testCustomTimeoutValues1() {
+        val config = PNConfiguration.builder(UserId(BasePubNubImpl.generateUUID()), "demo")
+        config.setPresenceTimeout(100)
```
marcin-cebo, line 35: Is this possible to do:
   val config = PNConfiguration.builder(UserId(BasePubNubImpl.generateUUID()), "demo")  
                                    .presenceTimeout(100)
                                    .heartbeatInterval(21)
                                    .retryConfiguration(new RetryConfiguraiton.Linear())
?

marcin-cebo (commented):
I didn't finish review. Just want to submit what I have so far.

pubnub-kotlin/src/main/kotlin/com/pubnub/api/v2/PNConfiguration.kt
```diff
@@ -0,0 +1,76 @@
...
+import okhttp3.CertificatePinner
+import okhttp3.ConnectionSpec
+import okhttp3.logging.HttpLoggingInterceptor
+import java.net.Proxy
+import java.net.ProxySelector
+import javax.net.ssl.HostnameVerifier
+import javax.net.ssl.SSLSocketFactory
+import javax.net.ssl.X509ExtendedTrustManager
+
+interface PNConfiguration : BasePNConfiguration {
```
marcin-cebo, line 20: In kotlin there are "named parameters". I thinking do we need builder patter for kotlin. What advantages has this approach over constructor with "named parameters"?

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/v2/BasePNConfiguration.kt
wkal-pubnub: just to follow the convention of having interfaces in core-api and implementations in core-impl really...
it also has some positive sides, for example lets us be less strict about visibility modifiers in core-impl classes, for example: if we need something visible for testing in `*Impl` we can just make it public because it won't be visible to SDK users anyway. Another side effect is that the build is faster when changes happen in *Impl classes only, especially when we further split java and kotlin into api/impl

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
```diff
@@ -135,10 +135,12 @@ sealed class RetryConfiguration {
             internal fun createForTest(
                 minDelayInSec: Duration = MIN_DELAY.seconds,
                 maxDelayInSec: Duration = MAX_DELAY.seconds,
-                maxRetryNumber: Int = MAX_RETRIES_IN_EXPONENTIAL,
+                maxRetryNumber: Int = MAX_RETRIES,
                 excludedOperations: List<RetryableEndpointGroup> = emptyList(),
                 isInternal: Boolean = false,
             ): Exponential = Exponential(minDelayInSec, maxDelayInSec, maxRetryNumber, excludedOperations, isInternal)
+
+            const val MAX_RETRIES = 6
```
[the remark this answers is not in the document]
wkal-pubnub: see comment above

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/PNConfigurationImplTest.kt
wkal-pubnub: ah, so that is a good question - I followed the convention from the old PNConfiguration and prefixed the methods with `set` and allowed chaining, so you can do:
```
PNConfiguration.builder(UserId(BasePubNubImpl.generateUUID()), "demo")
.setPresenceTimeout(100)
.setHeartbeatInterval(21)
.setRetryConfiguration(new RetryConfiguraiton.Linear())
```

some of our other builders use setters without `set`
the question is which should we use?
I'm leaning towards setX for PNConfiguration.Builder because the old PNConfiguration had it, so it will be easy to copy/paste old code and simply change the class to Builder... but I'm open to thoughts

pubnub-kotlin/src/main/kotlin/com/pubnub/api/v2/PNConfiguration.kt
wkal-pubnub: there are still advantages sometimes, especially for long parameter lists:
* it allows the user to pass around the builder or gradually build up the required parameters, without having to store them in a long list of variables to then call the constructor once
* it lets us start building a small DSL for when we start introducing groups of parameters (e.g. httpOptions, keyOptions, cryptoOptions) - tbf this could be also done in the constructor, but becomes a bit unwieldy with nested calls...
* it lets us evolve the available options and have setters with custom logic, for example making a change to a constructor is usually a breaking change (unless you're adding a new option at the end of the parameter list), but deprecating an option and adding a new one is useful to have

tl;dr it gives both us and the user more flexibility in how to construct the final object

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/retry/RetryConfiguration.kt
```diff
@@ -7,8 +7,6 @@ import kotlin.time.Duration.Companion.seconds
 
 private const val MIN_DELAY = 2
 private const val MAX_DELAY = 150
-private const val MAX_RETRIES_IN_LINEAR = 10
```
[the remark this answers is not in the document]
marcin-cebo: I see. Thanks.

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/v2/BasePNConfiguration.kt
marcin-cebo: ok, thanks.

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/PNConfigurationImplTest.kt
marcin-cebo: I think in Java SDK we have convention without set e.g.
```
        pubNub.publish()
                .message(messagePayload)
                .channel(expectedChannel)
                .sync()
```

```
        pubnub.subscribe()
                .channels(Arrays.asList(channels))
                .withPresence()
                .execute();
```

I see that for old customers having setters without `set` might require more effort but for new customers API will be consistent I think.

What do you think?

wkal-pubnub: yeah it might make sense to make it consistent, I'll remove the `set` prefix

pubnub-kotlin/src/main/kotlin/com/pubnub/api/v2/PNConfiguration.kt
marcin-cebo: ok, thanks

pubnub-gson/src/test/kotlin/com/pubnub/internal/v2/PNConfigurationImplTest.kt
wkal-pubnub: done

marcin-cebo (approved):

PR #197, opened 2024-03-29
Bring back PubNub-Java pnsdk String

wkal-pubnub:


pubnub-gson/src/main/java/com/pubnub/internal/PubNubImpl.java
```diff
@@ -144,7 +144,7 @@ public class PubNubImpl extends BasePubNubImpl<
     private final @NotNull BasePNConfiguration configuration;
 
     public PubNubImpl(@NotNull BasePNConfiguration configuration) {
-        super(configuration);
+        super(configuration, "PubNub-Java");
```
marcin-cebo, line 147: We could introduce a constant here. What do you think?

pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/PubNubCore.kt
```diff
@@ -112,16 +112,8 @@ class PubNubCore internal constructor(
     val configuration: BasePNConfiguration,
     val listenerManager: ListenerManager,
     eventEnginesConf: EventEnginesConf = EventEnginesConf(),
+    private val pnsdkName: String = "PubNub-JVM",
```
marcin-cebo, line 115: Why there is default value here?

marcin-cebo (approved):

pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/PubNubCore.kt
wkal-pubnub: I thought it might be useful to create PNCore in tests, but turns out I didn't use it. Removing

pubnub-gson/src/main/java/com/pubnub/internal/PubNubImpl.java
wkal-pubnub: will do

PR #198, opened 2024-03-29
Mc move retry

marcin-cebo:
This PR covers three commits the first one contains changes already reviewed in: 
https://github.com/pubnub/kotlin/pull/193
Second  and third commits contain additional changes to be review.


pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/eventengine/State.kt
```diff
@@ -14,6 +16,12 @@ internal fun <Ei : EffectInvocation, Ev : Event, S : State<Ei, Ev, S>> S.transit
     state: S,
     vararg invocations: Ei,
 ): Pair<S, Set<Ei>> {
+    val logger = LoggerFactory.getLogger(this::class.java)
```
wkal-pubnub, line 19: could we not create a logger every time this function is called?
maybe move it to the State ?

wkal-pubnub: unless `getLogger` doesn't allocate new objects... but I can't find any information on this

marcin-cebo: Good point. I will change it.

marcin-cebo: As it comes to: “This PR LGTM, but overall I think it's not great to tie together retry policy for EE and for one-off actions, especially related to publishing (where it can be desirable to fail fast as opposed to a long-poll loop). I'd like us to think of a way to make it more flexible, for example override retry per-endpoint, or per call (in the future)”

Currently there is a way to exclude group of API from retryConfiguration using excludedOperations pram
class Linear private constructor(
    var delayInSec: Duration = MIN_DELAY.seconds,
    var maxRetryNumber: Int = MAX_RETRIES,
    val excludedOperations: List<RetryableEndpointGroup> = emptyList(),

We could introduce retryConfig per call in the future. I could imagine it would be not for all API/endpoint but only those that we have good reason to do so.

PR #203, opened 2024-04-04
Split Java and Kotlin modules to API/impl modules

wkal-pubnub:


pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/PubNub.kt
```diff
@@ -571,7 +570,9 @@ interface PubNub :
          */
         @JvmStatic
         fun create(configuration: BasePNConfiguration): PubNub {
-            return PubNubImpl(configuration)
+            return Class.forName(
+                "com.pubnub.internal.PubNubImpl",
+            ).getConstructor(BasePNConfiguration::class.java).newInstance(configuration) as PubNub
```
marcin-cebo, line 575: Why reflection is needed here and in some other places?

marcin-cebo (approved):

PR #207, opened 2024-04-09
Add new options to config override for requests

wkal-pubnub:
feat: Added methods for publishing messages and signals on the `Channel` class. Also added new builder factories on `PubNub` with required parameters provided upfront, e.g. `PubNub.publish(message,channel)`. 

feat: Allow overriding certain PubNub configuration options per API call through `Endpoint.overrideConfiguration`

refactor: Remaining classes from `com.pubnub.internal` package were hidden from compilation classpath. Users should only use classes in `com.pubnub.api` package.


pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/v2/BasePNConfiguration.kt
```diff
@@ -581,5 +581,26 @@ interface BasePNConfigurationOverride {
...
+         * The value is in seconds.
+         *
+         * Defaults to 5.
+         */
+        val connectTimeout: Int
+
+        /**
+         * For non subscribe operations (publish, herenow, etc),
+         * This property relates to a read timeout that is applied from the moment the connection between a client
+         * and the server has been successfully established. It defines a maximum time of inactivity between two
```
marcin-cebo, line 597: This is not critical for now.

Currently we have property name "nonSubscribeRequestTimeout" that is a bit misleading because name suggests that this property relates to requestTimeout that can be understood like the maximum time a client will wait for a response from a server after it sends a request. This includes:
-resolving the hostname
-establishing a TCP connection
-TLS handsake
-sending the request data
-receiving the entire response data

In fact in our SDK "nonSubscribeRequestTimeout" controls read timeout that is applied from the moment the connection between a client and a server has been successfully established. It defines a maximum time of inactivity between two data packets when waiting for the servers's response.

It would be good to deprecate "nonSubscribeRequestTimeout" and introduce something like "nonSubscribeReadTimeout" instead.

This is not critical for now. Just want to share my observation.

marcin-cebo (approved):

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/v2/BasePNConfiguration.kt
wkal-pubnub: done

wkal-pubnub:
@pubnub-release-bot release kotlin as v9.1.0

PR #211, opened 2024-05-08
Replaced random with SecureRandom.

marcin-cebo:


pubnub-core/pubnub-core-impl/src/main/java/com/pubnub/internal/vendor/Crypto.java
```diff
@@ -50,7 +50,7 @@ public void initCiphers() throws PubNubException {
                     .toLowerCase().getBytes(ENCODING_UTF_8);
             if (dynamicIV) {
                 ivBytes = new byte[16];
-                new Random().nextBytes(ivBytes);
+                new SecureRandom().nextBytes(ivBytes);
```
wkal-pubnub, line 53: SecureRandom can be significantly slower than Random, so it would be nice to initialise it only once and store it for future calls

marcin-cebo: Ok, I will change it.

PR #215, opened 2024-05-20
Continue adding JS implementations for KMP

wkal-pubnub:
* changed Any? to Map<String,Any?>? on some PN* types to match the resulting data more closely
* moved some factory methods to separate files
* common tests now work on iOS. Need to have a simulator with id "iPhone 15 Pro" or edit gradle.properties and supply own simulator ID


build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubIosSimulatorTestPlugin.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.gradle.api.Project
+import org.gradle.api.tasks.Exec
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.register
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
+import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeSimulatorTest
+import java.io.ByteArrayOutputStream
+
+class PubNubIosSimulatorTestPlugin : Plugin<Project> {
```
marcin-cebo, line 13: What was the reason for creating this plugin?
My understanding is that available plugins (e.g. https://github.com/openbakery/gradle-xcodePlugin) don't fit our needs?

marcin-cebo (approved):

build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubIosSimulatorTestPlugin.kt
wkal-pubnub: I think the plugin you linked is about something else entirely - the readme says it's for building xcode projects I think? (it's not something we need here)

The reason is basically this: https://youtrack.jetbrains.com/issue/KT-54776

TL;DR: the Kotlin Multiplatform plugin tries to run iOS tests on the iOS Simulator with the `--standalone` flag by default, which for some reason makes all SSL requests fail due to missing certificates. 

I couldn't find any way to fix this other than the suggested workaround of disabling the `--standalone` flag, but then you need a mechanism to `boot` and `shutdown` the simulator before and after the test (the `standalone` flag took care of that but now it's disabled). 

So, the plugin introduces two new tasks that run before and after the iOS tests and make sure the simulator is running and shuts down after.

marcin-cebo: It's generated by ChatGpt4o

marcin-cebo (approved):

PR #228, opened 2024-06-06
Add creation of JsonElement to common code

wkal-pubnub:


pubnub-core/pubnub-core-api/src/jsMain/kotlin/com/pubnub/api/JsonElement.js.kt
```diff
@@ -34,10 +39,56 @@ actual fun JsonElement.asBoolean(): Boolean? {
...
 
 actual fun JsonElement.asNumber(): Number? {
     return value as? Number
+}
+
+actual fun createJsonElement(any: Any?): JsonElement {
+    return JsonElementImpl(any.adjustCollectionTypes())
+}
+
+internal fun Any?.adjustCollectionTypes(): Any? {
```
marcin-cebo, line 53: Did you consider using "kotlinx-serialization-json" lib? In such case it might be possible to make this method shorter like:

actual typealias JsonElement = kotlinx.serialization.json.JsonElement

actual fun JsonElement.asString(): String? {
    return (this as? JsonPrimitive)?.contentOrNull
}

actual fun JsonElement.asDouble(): Double? {
    return (this as? JsonPrimitive)?.doubleOrNull
}

actual fun JsonElement.asNumber(): Number? {
    return (this as? JsonPrimitive)?.intOrNull ?: (this as? JsonPrimitive)?.doubleOrNull
}
...
....
actual fun createJsonElement(any: Any?): JsonElement {
    return Json.encodeToJsonElement(any)
}

wkal-pubnub: our objects coming from the JVM SDK are of type `com.google.gson.JsonElement`, so doing `actual typealias JsonElement = kotlinx.serialization.json.JsonElement` would require changing the entire codebase of the JVM SDK to use kotlinx.serialization instead

wkal-pubnub: (which is something we would do in the future, potentially, if we unify the KMP SDK codebases)

marcin-cebo: Got it. Thanks.

marcin-cebo (approved):

pubnub-core/pubnub-core-api/src/jsMain/kotlin/com/pubnub/api/JsonElement.js.kt
wkal-pubnub: oh, sorry this is about JS, not JVM!
yes, that would make it a little bit easier, but I didn't want to take a dependency on kotlinx-serialization-json... I'll consider it though if maintaining this manually becomes too much work

PR #232, opened 2024-06-10
Enforce string-only values in data

wkal-pubnub:
feat: Add new FCMPayloadV2 for required new FCM push message format

fix: Security hardening: disallow DTD in XML parser and enable SecureRandom


marcin-cebo (approved):

pubnub-core/pubnub-core-api/src/main/kotlin/com/pubnub/api/models/consumer/push/payload/PushPayloadHelper.kt
```diff
@@ -126,6 +140,44 @@ class PushPayloadHelper {
...
+                        if (this.isNotEmpty()) {
+                            put("notification", this)
+                        }
+                    }
+                }
+            }
+        }
+
+        class Notification : PushPayloadSerializer {
+            var title: String? = null
```
marcin-cebo, line 166: It seem it could be extracted since there is a duplication. Notification is defined in FCMPayload and FCMPayloadV2 
?

wkal-pubnub: It's intentional - I can't move the `FCMPayload.Notification` class because it would be a breaking change, and I can't use it in `FCMPayloadV2` because `FCMPayload` will be deleted in the future. And in any case after the removal of the deprecated FCMPayload (v1), only FCMPayloadV2.Notification will remain.

marcin-cebo: Ok

wkal-pubnub:
@pubnub-release-bot release kotlin as v9.2.0

wkal-pubnub:
@pubnub-release-bot release kotlin as v9.2.0

PR #239, opened 2024-06-18
File download + File upload

jguz-pubnub:


pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/endpoints/files/SendFile.ios.kt
```diff
@@ -1,9 +1,83 @@
...
+        mapContent(inputStream)?.let { content ->
+            pubnub.sendFileWithChannel(
+                channel = channel,
+                fileName = fileName,
+                message = message,
+                content = content,
+                meta = meta,
+                ttl = ttl?.let { NSNumber(it) },
+                shouldStore = shouldStore?.let { NSNumber(it) },
+                onSuccess = callback.onSuccessHandler2 { file, timetoken ->
```
marcin-cebo, line 51: FYI
There is file here and in line 55. I seems that they are two different object. Maybe we can rename one or the other or it doesn't make sense?

pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/endpoints/files/SendFile.ios.kt
```diff
@@ -1,9 +1,83 @@
...
+                    PNFileUploadResult(
+                        status = 200,
+                        timetoken = timetoken.toLong(),
+                        file = PNBaseFile(
+                            id = file?.id() ?: "",
```
marcin-cebo, line 56: If you like you could change it to:
id = file?id().orEmpty()

marcin-cebo (approved):

marcin-cebo (approved):

PR #240, opened 2024-06-19
Add IncludeType, fix futures.awaitAll and listener classcastexception on JS

wkal-pubnub:


pubnub-kotlin/pubnub-kotlin-test/build.gradle.kts
```diff
@@ -44,11 +43,11 @@ kotlin {
...
 //        }
 //
-//        val jvmMain by getting {
-//            dependencies {
-//                implementation(project(":pubnub-core:pubnub-core-impl"))
-//            }
-//        }
+        val jvmMain by getting {
+            dependencies {
+                api(kotlin("test-junit"))
```
marcin-cebo, line 48: I guess that test-junit is base on old JUnit4.
What about using JUnit5
   testImplementation("org.junit.jupiter:junit-jupiter-api:5.8.1")
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.8.1")

wkal-pubnub: I need the KMP dependency compatible with `kotlin-test` here, I can't use these other dependencies, and we're not accessing JUnit directly in our code, only through `kotlin-test`

pubnub-core/pubnub-core-impl/src/main/kotlin/com/pubnub/internal/PubNubCore.kt
```diff
@@ -1315,6 +1315,7 @@ class PubNubCore internal constructor(
         includeCount: Boolean = false,
         includeCustom: Boolean = false,
         includeChannelDetails: PNChannelDetailsLevel? = null,
+        includeType: Boolean = false,
```
marcin-cebo, line 1318: What is a reason we want to expose `includeType` ?

wkal-pubnub: it's needed for ChatSDK, otherwise returned data will always have a type UNKNOWN

besides, it's present in the REST API and so far it wasn't implemented in the SDK

marcin-cebo: I see. Thank's.

marcin-cebo (approved):

PR #242, opened 2024-06-24
Minor improvements

jguz-pubnub:
* Getting rid of warning related with Maps
* Ensuring that JsonImpl is used as expected
* Added some missing implementations in PubNubImpl.kt

### DIFF
--- a/pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/kmp/Converters.kt
+++ b/pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/kmp/Converters.kt
@@ -39,7 +39,7 @@ internal fun createPNUUIDMetadata(from: PubNubUUIDMetadataObjC): PNUUIDMetadata 
         externalId = from.externalId(),
         profileUrl = from.profileUrl(),
         email = from.email(),
-        custom = from.custom()?.asMap() as? Map<String, Any?>, // TODO: Verify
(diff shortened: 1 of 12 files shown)


pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/kmp/Converters.kt
```diff
@@ -95,13 +95,18 @@ internal fun createPNChannelMembership(from: PubNubMembershipMetadataObjC): PNCh
...
     )
 }
 
 internal inline fun <reified T, U> List<*>?.filterAndMap(mapper: (T) -> U): Collection<U> {
     return this?.filterIsInstance<T>()?.map(mapper) ?: emptyList()
+}
+
+@Suppress("UNCHECKED_CAST")
+inline fun <reified K, reified V> Map<*, *>.safeCast(): Map<K, V> {
+    return this as? Map<K,V> ?: emptyMap()
```
wkal-pubnub, line 111: I think at this stage I'd prefer if this was
`return this as? Map<K,V> ?: error("Cannot make the cast")`
so that we can easily find all places where this cast doesn't actually succeed when testing
I'm worried that if we silently return an empty map we might miss it

jguz-pubnub: Right, that makes sense. I'll fix it

wkal-pubnub (approved):

PR #251, opened 2024-07-04
SetState via Heartbeat fix.

marcin-cebo:
fix: SetState via Heartbeat fix


pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/endpoints/presence/SetStateImpl.java
```diff
@@ -86,7 +88,19 @@ protected EndpointInterface<Object> createAction() {
...
+            return (EndpointInterface<Object>) (Object) new HeartbeatEndpoint(pubnub, channels, channelGroups, composeStateParamValue());
         }
     }
+
+    private Object composeStateParamValue() {
+        Map<String, Object> stateParamValue = new HashMap<>();
+        for (String channel : channels) {
+            stateParamValue.put(channel, state);
+        }
+        for (String channelGroup : channelGroups) {
```
wkal-pubnub, line 100: only channels are allowed in the map, 

maybe we should even check it in `validateParams` that no channel groups where passed if `withHeartbeat` is `true`

wkal-pubnub (approved):

marcin-cebo:
@pubnub-release-bot release kotlin as v9.2.2

PR #253, opened 2024-07-09
Change actionTimetoken to Long, ktlint, tests

wkal-pubnub:


pubnub-kotlin/pubnub-kotlin-test/src/commonMain/kotlin/com.pubnub.test/BaseIntegrationTest.kt
```diff
@@ -52,17 +53,22 @@ abstract class BaseIntegrationTest {
...
+        async { result ->
+            result.onSuccess {
+                cont.resume(it)
+            }.onFailure {
+                cont.resumeWithException(it)
+            }
         }
     }
+    withContext(Dispatchers.Default) {
+        delay(100)
```
marcin-cebo, line 67: Why 100 ?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-test/src/commonMain/kotlin/com.pubnub.test/BaseIntegrationTest.kt
wkal-pubnub: I found that in general tests tend to still be flaky due to replication/caching on the server and adding even a small delay after each action (hence in this await function) made them much more reliable. 

In general I agree though that it's proving difficult to keep the integration tests working 100% due to the server not offering any kind of reliability of data transactions :( I'll have to rething this whole thing... again.

marcin-cebo: ok, thank you for clarification.

PR #255, opened 2024-07-16
Refactoring around PNConfiguration

wkal-pubnub:


pubnub-kotlin/pubnub-kotlin-api/src/jsMain/kotlin/com/pubnub/api/PubNubImpl.kt
```diff
@@ -1153,7 +1165,7 @@ fun PNConfiguration.toJs(): PubNubJs.PNConfiguration {
     config.subscribeKey = subscribeKey
     config.publishKey = publishKey
     config.secretKey = secretKey
-    config.logVerbosity = logVerbosity
+    config.logVerbosity = logVerbosity == PNLogVerbosity.BODY
```
marcin-cebo, line 1168: Do we want to have `PNLogVerbosity.BODY` as default ?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-api/src/jsMain/kotlin/com/pubnub/api/PubNubImpl.kt
wkal-pubnub: please note that this checks for equality `==` and not `=`, so it's only set to true when `BODY` is passed in arguments

PR #265, opened 2024-08-01
Build scripts with non-JVM targets disabled

wkal-pubnub:
New gradle.properties:
```
ENABLE_TARGET_JS=false
ENABLE_TARGET_IOS=false
ENABLE_TARGET_IOS_SIMULATOR=false
```

By default only JVM part is enabled for faster builds.


pubnub-kotlin/pubnub-kotlin-api/src/commonTest/kotlin/com/pubnub/test/integration/ChannelMetadataTest.kt
```diff
@@ -29,7 +30,7 @@ class ChannelMetadataTest : BaseIntegrationTest() {
     private val type = randomString()
 
     @Test
-    fun can_set_metadata() = runTest(timeout = defaultTimeout) {
```
marcin-cebo, line 32: Why timeout being set to default is not needed any more?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-api/src/commonTest/kotlin/com/pubnub/test/integration/ChannelMetadataTest.kt
wkal-pubnub: because I noticed that instead of remembering to add it to every single method, this can be controlled by setting the `kotlinx.coroutines.test.default_timeout` system property, so for example we can give higher timeout on CI if needed

PR #266, opened 2024-08-05
[Kotlin] Change PNChannelMetadata and PNUUIDMetadata to contain partial update information

wkal-pubnub:
The `PNChannelMetadata` and `PNUUIDMetadata` now represents information received from the server accurately - it contains `null` in properties that were not received (they were not updated), and `PatchValue(someValue)` or `PatchValue(null)` in properties that were part of the update.


pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
```diff
@@ -1,12 +1,29 @@
...
+    val custom: PatchValue<Map<String, Any?>?>?,
+    val updated: PatchValue<String>?,
+    val eTag: PatchValue<String>?,
+    val type: PatchValue<String?>?,
+    val status: PatchValue<String?>?,
+) {
+    /**
+     * Merge information from this `PNChannelMetadata` with new data from `update`, returning a new `PNChannelMetadata` instance.
+     */
+    operator fun plus(update: PNChannelMetadata): PNChannelMetadata {
```
marcin-cebo, line 18: For me name "plus" is misleading. What do you think about "applyUpdate" or "combineWith" ?
and then to retain plus operator code could be modified like this:

```
    operator fun plus(update: PNChannelMetadata): PNChannelMetadata{
       return applyUpdate(update)
    }

    fun applyUpdate(update: PNChannelMetadata): PNChannelMetadata {
        return copy(
            name = update.name ?: name,
            description = update.description ?: description,
            custom = update.custom ?: custom,
            updated = update.updated ?: updated,
            eTag = update.eTag ?: eTag,
            type = update.type ?: type,
            status = update.status ?: status,
        )
    }
```

pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
```diff
@@ -1,12 +1,29 @@
...
+    val name: PatchValue<String?>?,
+    val description: PatchValue<String?>?,
+    val custom: PatchValue<Map<String, Any?>?>?,
+    val updated: PatchValue<String>?,
+    val eTag: PatchValue<String>?,
+    val type: PatchValue<String?>?,
+    val status: PatchValue<String?>?,
+) {
+    /**
+     * Merge information from this `PNChannelMetadata` with new data from `update`, returning a new `PNChannelMetadata` instance.
```
marcin-cebo, line 16: What about adding unit test for this. Does it make sense?

pubnub-kotlin/pubnub-kotlin-impl/src/integrationTest/kotlin/com/pubnub/api/integration/HistoryIntegrationTest.kt
```diff
@@ -176,7 +176,7 @@ class HistoryIntegrationTest : BaseIntegrationTest() {
                                 expectedActionValue to
                                     listOf(
                                         PNFetchMessageItem.Action(
-                                            actionTimetoken = actionResult.actionTimetoken.toString(),
+                                            actionTimetoken = actionResult.actionTimetoken!!,
```
marcin-cebo, line 179: Do we have integration test that checks if `PatchValue` works as expected?
I am thinking about full scenario: 
createMetadata, 
updateMetadata, 
validate some value has been set to null
validate some value has been remove
validate some value has been added

Then when different part of functionality(sending to server) will be done such test could be just lightly modified.
What do you think?

pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
wkal-pubnub: regarding `applyUpdate` - it suggests that the object will change which is not the case, and regarding `combineWith` IMHO it literally means `plus`, so I don't really like any of the options, nor do I think that we need to have two functions doing the same thing in Kotlin, especially since the developer is supposed to do `a + b` and not use `plus` directly. However, the question may come up in the future when we expose this to Java, so I'm open to other suggestions for names, but so far I think it's not an improvement over `plus` (which is the required name for the operator override in Kotlin)

pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
wkal-pubnub: yeah, I forgot to do it. Will add

pubnub-kotlin/pubnub-kotlin-impl/src/integrationTest/kotlin/com/pubnub/api/integration/HistoryIntegrationTest.kt
wkal-pubnub: Yes, I modified existing integration tests in two places to make sure that the data is read correctly from the server. I also have a unit test that checks that deserialization works correctly just for reading PatchValue from JSON.

But you're right that maybe once we have sending to the server, we should check the whole "journey" of the data to server and back explicitly. Will add a todo

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/v2/entities/Channel.kt
```diff
@@ -110,9 +110,9 @@ interface Channel : BaseChannel<EventListener, Subscription> {
      */
     fun fire(
         message: Any,
-        meta: Any?,
-        usePost: Boolean,
-        ttl: Int?
+        meta: Any? = null,
```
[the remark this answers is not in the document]
wkal-pubnub: I can look into it and add a deprecation on TTL if possible

pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
marcin-cebo: For me this is not really "plus" operation but rather updateIfApplicable or just update.
oldDescritpion = "Old Desc"
newDescription = "New Desc" 

oldDescritpion + newDescription = "Old Desc New Desc"
oldDescritpion.updateIfApplicable(newDescription) = "New Desc"

wkal-pubnub: the plus operation can be many things depending on the domain: 
for a vector it can be what you described: (a,b,c) + (x,y,z) = (x+a, y+b, z+c)
for two lists it's [a] + [b, c] = [a, b, c] , (and not [ab, c])
for two maps it's {"a": "abc"} + {"a": "xyz"} = {"a": "xyz"} (right one wins)

for a change set (a data class containing what I called PatchValues) like we have here it's most similar two adding two maps, which can still be described as a plus operation

wkal-pubnub: what I'm saying is that the `plus` operator in Kotlin isn't always used for arithmetic plus or concatenation

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/v2/entities/Channel.kt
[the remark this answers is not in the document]
wkal-pubnub: done, added new method and deprecated old one

pubnub-core/pubnub-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/channel/PNChannelMetadata.kt
marcin-cebo: I see. Ok.

marcin-cebo (approved):

PR #270, opened 2024-08-16
Fix crash on Android in retry logic after destroy called

wkal-pubnub:
fix: Fixes a crash on Android after `PubNub.destroy` is called and there are requests running


marcin-cebo (approved):

wkal-pubnub:
@pubnub-release-bot release kotlin as v9.2.4

PR #276, opened 2024-09-10
Fixes according to swift module changes

jguz-pubnub:
- Fixed import statements in iosMain due to names refactor (adding the `KMP` prefix) in the `swift` submodule.
- Made changes in the `swift` submodule, replacing the product name with `PubNubSDK` due to a compiler error that prevents emitting the module interface if one of the module's classes has the same name as the module itself
- Changed module name in `PubNubKotlinMultiplatformPlugin.kt`

### DIFF
--- a/build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubKotlinMultiplatformPlugin.kt
+++ b/build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubKotlinMultiplatformPlugin.kt
@@ -111,7 +111,7 @@ class PubNubKotlinMultiplatformPlugin : Plugin<Project> {
 //                        }
 //                        version = "7.1.0"
                             source = path(rootProject.file(swiftPath))
-                            moduleName = "PubNub"
+                            moduleName = "PubNubSDK"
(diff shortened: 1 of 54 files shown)


build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubKotlinMultiplatformPlugin.kt
```diff
@@ -111,7 +111,7 @@ class PubNubKotlinMultiplatformPlugin : Plugin<Project> {
 //                        }
 //                        version = "7.1.0"
                             source = path(rootProject.file(swiftPath))
-                            moduleName = "PubNub"
+                            moduleName = "PubNubSDK"
```
marcin-cebo, line 114: What is a reason for this change?

marcin-cebo: I found :)

> Made changes in the swift submodule, replacing the product name with PubNubSDK due to a compiler error that prevents emitting the module interface if one of the module's classes has the same name as the module itself

marcin-cebo (approved):

PR #280, opened 2024-09-19
Added @ObjCName annotation and helper constructor (iOS)

jguz-pubnub:
- Added `@ObjCName` annotation due to `PubNub` name ambiguity. It's present both in KMP binary and Swift SDK. This fix prevents us from adding prefixes in the Swift chat layer 
- Added companion object for `PubNubImpl` in iOS. Xcode does not allow me to call a default constructor and it can't infer that the `KMPPubNub` type belongs to Swift Chat SDK. Fortunately, this change is completely hidden from any customer

### DIFF
--- a/pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
+++ b/pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
@@ -152,6 +152,12 @@ class PubNubImpl(private val pubNubObjC: KMPPubNub) : PubNub {
         )
     )
 
+    companion object {
+        fun create(kmpPubNub: Any): PubNubImpl {
+            return PubNubImpl(kmpPubNub as KMPPubNub)
+        }
+    }
+
     override val configuration: PNConfiguration = createPNConfiguration(
         UserId(pubNubObjC.configObjC().userId()),
         "", // todo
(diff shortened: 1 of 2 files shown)


pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
```diff
@@ -152,6 +152,12 @@ class PubNubImpl(private val pubNubObjC: KMPPubNub) : PubNub {
         )
     )
 
+    companion object {
+        fun create(kmpPubNub: Any): PubNubImpl {
```
wkal-pubnub, line 156: It seems weird that you can't just call the constructor... this should work I think.
Anyway, this `create` is in iOS only so it's fine, but I'd be curious to see what the error is with the constructor, maybe we can fix it

wkal-pubnub (approved):

pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
jguz-pubnub: Yeah, it' really weird. I see the original method I want to invoke is present in KMP binary:

pubnubimpl_init

However, here's the error when I want to invoke it. It says `argument passed to call that takes no arguments`:

pubnubimpl_error_usage

PR #287, opened 2024-09-25
Fixed dependecy definition in README

marcin-cebo:


README.md
```diff
@@ -26,7 +26,7 @@ You will need the publish and subscribe keys to authenticate your app. Get your
 
    * for Gradle, add the following dependency in your `gradle.build`:
```
wkal-pubnub, line 27: if we're changing it to the Kotlinscript format (which is recommended) in that case this should be `gradle.build.kts`

wkal-pubnub (approved):

PR #291, opened 2024-10-01
Added util object to handle time

marcin-cebo:


pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Converts a PubNub timetoken (a unique identifier for each message sent and received in a PubNub channel that is
+     * a number of 100-nanosecond intervals since January 1, 1970) to LocalDateTime object representing
+     * the corresponding date and time.
+     *
+     * @param timetoken PubNub timetoken
+     * @return [LocalDateTime] representing the corresponding data and time in UTC.
+     * @throws IllegalArgumentException if the timetoken does not have 17 digits.
+     */
+    fun timetokenToDateTimeUTC(timetoken: Long): LocalDateTime {
```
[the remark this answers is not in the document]
marcin-cebo: Ok, I will make a change.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Converts [LocalDateTime] in UTC to a PubNub timetoken
+     *
+     * A PubNub timetoken is a 17-digit number representing the number of 100-nanosecond intervals since January 1, 1970.
+     *
+     * @param dateTime The [LocalDateTime] object in UTC to be converted to a PubNub timetoken.
+     * @return A 17-digit [Long] representing the PubNub timetoken for the given [LocalDateTime].
+     */
+    fun dateTimeUTCToTimetoken(dataTime: LocalDateTime): Long {
```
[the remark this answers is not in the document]
marcin-cebo: Ok, I will make a change.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,93 @@
...
+     * Converts a Unix timestamp (in seconds)to a PubNub timetoken
+     *
+     * A Unix timestamp represents the number of seconds that have elapsed since January 1, 1970 (midnight UTC/GMT).
+     * This function converts the Unix timestamp, which must be provided in seconds, to a PubNub timetoken.
+     *
+     * @param unixTime The Unix timestamp in seconds to be converted to a PubNub timetoken.
+     * @return A 17-digit [Long] representing the PubNub timetoken corresponding to the given Unix timestamp.
+     */
+    fun unixToTimetoken(unixTime: Long): Long {
+        return (unixTime * 10_000_000)
```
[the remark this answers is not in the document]
marcin-cebo: Ok, I will make a change.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * A PubNub timetoken is a 17-digit number representing the number of 100-nanosecond intervals since January 1, 1970.
+     * This function converts the PubNub timetoken to a Unix timestamp by reducing the precision to seconds.
+     * Note that precision finer than seconds is lost in this conversion.
+     *
+     * @param timetoken The PubNub timetoken to be converted to a Unix timestamp.
+     * @return A [Long] representing the Unix timestamp in seconds corresponding to the given timetoken.
+     */
+    fun timetokenToUnix(timetoken: Long): Long {
+        return (timetoken / 10_000_000)
```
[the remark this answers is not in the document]
marcin-cebo: Ok, I will make a change.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Converts a Unix timestamp to a PubNub timetoken
+     *
+     * @param unixTime The Unix timestamp to be converted to a PubNub timetoken.
+     * @return A 17-digit [Long] representing the PubNub timetoken corresponding to the given Unix timestamp.
+     */
+    fun unixToTimetoken(unixTime: Long): Long {
+        val numberOfDigits = unixTime.toString().length
```
wkal-pubnub, line 67: why not just `= unixTime * 10_000` ?

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,89 @@
...
+        val pnTimetokenInNanosecondsWithoutNanoPrecision = epochSeconds * 10_000_000
+        val nanosecondsForPubNubTimetoken =
+            epochNanoseconds / 100 // PubNub timetoken for nanoseconds store only 7 digits instead of 9
+        return pnTimetokenInNanosecondsWithoutNanoPrecision + nanosecondsForPubNubTimetoken
+    }
+
+    /**
+     * Converts a Unix timestamp to a PubNub timetoken
+     *
+     * @param unixTime The Unix timestamp to be converted to a PubNub timetoken.
```
wkal-pubnub, line 63: this documentation doesn't mention that the unixTime must be in milliseconds

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
wkal-pubnub: as of now this test won't pass:

```
    @Test
    fun conversion_is_reversible() {
        val unixTimes = listOf(10, 100, 1000, 10_000, 100_000, 1_000_000_000_000)
        unixTimes.forEach {
            assertEquals(it, TimetokenUtil.timetokenToUnix(TimetokenUtil.unixToTimetoken(it)))
        }
    }

```

marcin-cebo: I added validation and description that unix timetoken should has 10 or 13 digits.

Conversion:
assertEquals(it, TimetokenUtil.timetokenToUnix(TimetokenUtil.unixToTimetoken(it)))
is reversible only when unix timetoken has 13 digits because TimetokenUtil.timetokenToUnix returns 13 digits.

What do you think?

wkal-pubnub: then if we want to have this limitation that we need 13 digits, I think the `unixToTimetoken` function should **only** accept milliseconds.

As it is, the function which could be just `* 10000` is way too complicated... it's using string conversions, counting characters, doubles, pow() and so on... and all that to handle milliseconds or seconds, which IMHO makes it just more confusing. 

I would keep it simple both from the perspective of what the function does (i.e. converts milliseconds to timetokens) as well as the implementation side. If you think we should have a function that takes seconds, it should be a separate function, but IMHO it's not needed

marcin-cebo: What do you think about:
```
    fun unixToTimetoken(unixTime: Long): Long {
        val unixTimeLength = unixTime.toString().length

        return when (unixTimeLength) {
            UNIX_TIME_IN_SECONDS -> { unixTime * 1_000_000 }
            UNIX_TIME_IN_MILLISECONDS -> { unixTime * 1_000 }
            else -> { throw IllegalArgumentException("Unix timetoken should have 10 or 13 digits") }
        }
    }
```

wkal-pubnub: this only resolves the concerns about the implementation (though the string conversion is still unnecessary, I'd recommend just comparing MIN > unixTime > MAX to see if it's in the valid range)

but honestly I don't think this function can take two different units and magically return the correct result, because there's no way to fix the reverse function - `timetokenToUnix` cannot guess which unit you want to get back. 

The only way this could work with seconds AND milliseconds (and any other unit) is by adding a `unit: DurationUnit` parameter to both `unixToTimetoken` and `timetokenToUnix` and then the user can pass whichever unit they expect.

Otherwise this is just too confusing and error-prone

wkal-pubnub: for example: I have a class `MyTimestampUtility` with methods like `saveTimestamp()` and `getTimestamp()` . It operates on *seconds* (and not *milliseconds*).

I use `getTimestamp()` in seconds, then pass it to `getHistory(from = unixToTimetoken(seconds))`, and I get some messages back. Now I select the last message I'm interested in and want to save its timestamp back to my object.
So naturally I do `myTimestampUtility.saveTimestamp( timetokenToUnix(message.timetoken) )`... which is wrong

can you see  why I really don't like it?

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
marcin-cebo: ok, I will add a unit

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
marcin-cebo: I think I see your point. For me the fact that `timetokenToUnix` returns only milliseconds is fine, this fact is documented. Removing seconds handling from `unixToTimetoken` cause that users operating on seconds have more work todo.

I will remove seconds handling from `unixToTimetoken` method.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/util/TimetokenUtil.java
```diff
@@ -0,0 +1,95 @@
+package com.pubnub.api.java.util;
+
+import java.time.Instant;
+
+public class TimetokenUtil {
```
wkal-pubnub, line 5: nit: even if we're writing a separate implementation, this class could have been written in Kotlin, we don't need to write Java code in `pubnub-gson`

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/utils/TimetokenUtil.kt
```diff
@@ -0,0 +1,101 @@
...
+        private const val MAXIMUM_UNIX_TIME_VALUE = 9_999_999_999_999
+        private const val UNIX_TIME_LENGHT_WHEN_IN_MILLISECONDS = 13
+
+        /**
+         * Converts a PubNub timetoken (a unique identifier for each message sent and received in a PubNub channel that is
+         * a number of 100-nanosecond intervals since January 1, 1970) to LocalDateTime object representing
+         * the corresponding date and time.
+         *
+         * @param timetoken PubNub timetoken
+         * @return [Instant] representing the corresponding data and time.
```
wkal-pubnub, line 28:
```suggestion
         * @return [Instant] representing the corresponding moment in time.
```

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/util/TimetokenUtil.java
```diff
@@ -0,0 +1,95 @@
...
+        throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
+    }
+
+    /**
+     * Converts a PubNub timetoken (a unique identifier for each message sent and received in a PubNub channel that is
+     * a number of 100-nanosecond intervals since January 1, 1970) to Instant object representing
+     * the corresponding date and time.
+     *
+     * @param timetoken PubNub timetoken
+     * @return Instant representing the corresponding date and time.
```
wkal-pubnub, line 26:
```suggestion
     * @return Instant representing the corresponding moment in time.
```

wkal-pubnub (approved):

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/util/TimetokenUtil.java
marcin-cebo: I like this suggestion! Thanks. I will rewrite it.

wkal-pubnub (approved):

PR #293, opened 2024-10-03
Fix and enable JS build

wkal-pubnub:


pubnub-kotlin/pubnub-kotlin-api/src/jsMain/kotlin/com/pubnub/kmp/converters.js.kt
```diff
@@ -26,33 +27,33 @@ internal fun SetChannelMetadataResponse.toChannelMetadataResult(): PNChannelMeta
...
-        uuid.name,
-        uuid.externalId,
-        uuid.profileUrl,
-        uuid.email,
-        uuid.custom?.toMap(),
-        uuid.updated,
-        uuid.eTag,
-        uuid.type,
-        uuid.status
+        PatchValue.of(uuid.name), // TODO support optionals
```
marcin-cebo, line 43: What does it mean "support optionals" here?

pubnub-kotlin/pubnub-kotlin-api/src/jsMain/kotlin/com/pubnub/kmp/factories.js.kt
```diff
@@ -165,10 +166,10 @@ actual fun createEventListener(
                                 PNSetMembershipEvent(
                                     it.channel.id,
                                     it.uuid.id,
-                                    it.custom?.toMap(),
+                                    PatchValue.of(it.custom?.toMap()),
                                     it.eTag,
                                     it.updated,
-                                    null // todo missing
+                                    PatchValue.of(it.asDynamic().status) // todo missing
```
marcin-cebo, line 172: Could todo be removed?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-api/src/jsMain/kotlin/com/pubnub/kmp/converters.js.kt
wkal-pubnub: right now it just takes nulls and puts them in PatchValue so it's not really optional, it's just always a value (which can be null)

I have to see if there's a way to get the information from JS SDK about the presence (or absence) of value

PR #296, opened 2024-10-04
Switch runner

wkal-pubnub:


.github/workflows/release.yml
```diff
@@ -10,10 +10,15 @@ jobs:
     name: Check release required
     if: github.event.pull_request.merged && endsWith(github.repository, '-private') != true
     runs-on:
-      group: Default
+      group: macos-gh
     outputs:
       release: ${{ steps.check.outputs.ready }}
     steps:
+      - name: Install JDK
```
marcin-cebo, line 17: I am wondering why we didn't have this step before and now is needed?

marcin-cebo (approved):

.github/workflows/release.yml
```diff
@@ -10,10 +10,15 @@ jobs:
     name: Check release required
     if: github.event.pull_request.merged && endsWith(github.repository, '-private') != true
     runs-on:
-      group: Default
+      group: macos-gh
```
jguz-pubnub, line 13: I assume `macos-gh` has Xcode 15.3 or 15.4 installed, so we shouldn't encounter problems while building iOS target. Let's see it in action

.github/workflows/release.yml
wkal-pubnub: I don't know, maybe the default GH runner has Java preinstalled, but our mac os one doesn't

PR #299, opened 2024-10-11
Added getToken to Kotlin SDK

marcin-cebo:
so that it can be used in Chat KMP.

There is no need to add getToken to Java.

Deprecated authKey config property.

### DIFF
--- a/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
+++ b/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
@@ -713,6 +713,9 @@ interface PubNub : EventEmitter, StatusEmitter {
     @Throws(PubNubException::class)
     fun parseToken(token: String): PNToken
 
+    /**
+     * Method use in client devices to update the authenticated token granted by the server.
+     */
     fun setToken(token: String?)
 
     /**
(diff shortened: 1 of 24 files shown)


pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -713,6 +713,9 @@ interface PubNub : EventEmitter, StatusEmitter {
     @Throws(PubNubException::class)
     fun parseToken(token: String): PNToken
 
+    /**
+     * Method use in client devices to update the authenticated token granted by the server.
```
wkal-pubnub, line 717:
```suggestion
     * Update the authorization token granted by the server.
```

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/v2/PNConfiguration.kt
```diff
@@ -59,8 +59,11 @@ interface PNConfiguration : com.pubnub.api.v2.PNConfiguration {
         val secretKey: String
 
         /**
-         * If Access Manager is utilized, client will use this authKey in all restricted requests.
+         * If Access Manager (deprecated PAM v2) is utilized, client will use this authKey in all restricted requests.
          */
+        @Deprecated(
+            "This setting is deprecated because it relates to deprecated Access Manager (PAM V2). Use method pubnub.setToken(token) instead.",
```
wkal-pubnub, line 65: IMHO in all the new documentation strings we're missing the opportunity to educate the user about the difference between v2 and v3. In all your changes in this PR, the call to action for the user is to take the PAM v2 authKey and set it through pubnub.setToken. But in reality, they should migrate from v2 to v3 by going to https://www.pubnub.com/docs/general/resources/migration-guides/pam-v3-migration and making the appropriate changes to their infrastructure. I think we should link to the guide and be more explicit about using v3 token instead of authKey

PR #300, opened 2024-10-16
Add authToken property in PubNubImpl (iOS)

jguz-pubnub:


kotlin-js-store/yarn.lock
jguz-pubnub, line 1: This is the change I have after running `./gradlew build`. Should I commit it or ignore for now?

build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubKotlinMultiplatformPlugin.kt
```diff
@@ -41,7 +41,7 @@ class PubNubKotlinMultiplatformPlugin : Plugin<Project> {
                         pod("PubNubSwift") {
 //                            val swiftPath = project.findProperty("SWIFT_PATH") as? String ?: "swift"
 //                            source = path(rootProject.file(swiftPath))
-                            version = "8.0.0"
+                            source = git("https://github.com/pubnub/swift") { source -> source.branch = "feat/kmp-auth-token" }
```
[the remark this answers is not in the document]
jguz-pubnub: However, the problem I see is a scenario where we don't have anything scheduled to release in Swift SDK, but there are some bugs/improvements in the Objective-C compatible layer. Then, we're forced to release a new Swift SDK with some version that contains changes for our internal usage only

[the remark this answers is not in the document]
wkal-pubnub: there's no other way, when we want to release a PubNub KMP version that depends on some changes in PubNub Swift SDK, that new PubNub Swift SDK must be public
we can't use `source = ...` or our users would have to also build their app with that Swift version from a branch
we have to use `version = ... `

kotlin-js-store/yarn.lock
wkal-pubnub: it's fine to commit

pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
```diff
@@ -818,8 +818,7 @@ class PubNubImpl(private val pubNubObjC: KMPPubNub) : PubNub {
     }
 
     override fun getToken(): String? {
-        // TODO:
-        return null
+        return pubNubObjC.configObjC().authToken()
```
wkal-pubnub, line 821: so in swift when someone calls `pubnub.setToken` it is saved in `pubnub.config.authToken` ?

wkal-pubnub (dismissed):

build-logic/gradle-plugins/src/main/kotlin/com/pubnub/gradle/PubNubKotlinMultiplatformPlugin.kt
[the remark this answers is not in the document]
wkal-pubnub: actually, don't merge this please
merge with correct version 8.0.1

pubnub-kotlin/pubnub-kotlin-api/src/iosMain/kotlin/com/pubnub/api/PubNubImpl.kt
jguz-pubnub: I agreed with Marcin that I'll change this getter to method. Te final invocation should look like:

```
return pubNubObjC.configObjC().getToken()
```

marcin-cebo (approved):

wkal-pubnub (approved):

marcin-cebo (approved):

PR #305, opened 2024-11-05
Added authToken to kotlin PNConfiguration

marcin-cebo:
Now it is possible in kotlin for client-side PubNub initialization to provide authToken that will be used in API calls to server. It is also possible in kotlin to override configuration with new token and use it to call particular API e.g. pubnub.publish(randomChannel(), "message").overrideConfiguration {
    publishKey = Keys.pubKey
    authToken = expectedAuthToken
}.sync()

### DIFF
--- a/pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
+++ b/pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
@@ -25,6 +25,7 @@ class PNConfigurationImpl(
     override val publishKey: String = "",
     override val secretKey: String = "",
     override val authKey: String = "",
+    override val authToken: String = "", // this property is not used
     override val cryptoModule: CryptoModule? = null,
     override val origin: String = "",
     override val secure: Boolean = true,
--- a/pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
+++ b/pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
@@ -25,7 +25,7 @@ class PNConfigurationImpl(
     override val publishKey: String = "",
     override val secretKey: String = "",
     override val authKey: String = "",
-    override val authToken: String = "", // this property is not used
+    override val authToken: String? = null, // this property is not used, user can't create configuration with authToken
     override val cryptoModule: CryptoModule? = null,
     override val origin: String = "",
     override val secure: Boolean = true,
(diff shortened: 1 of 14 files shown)


pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/v2/PNConfigurationImpl.kt
```diff
@@ -106,6 +107,12 @@ class PNConfigurationImpl(
         )
         override var authKey: String = defaultConfiguration.authKey
 
+        /**
+         * Authentication token for the PubNub client. This token is required on the client side when Access Manager (PAM) is enabled for PubNub keys.
```
wkal-pubnub, line 111: no KDoc on `override`s

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
```diff
@@ -25,6 +25,7 @@ class PNConfigurationImpl(
     override val publishKey: String = "",
     override val secretKey: String = "",
     override val authKey: String = "",
+    override val authToken: String = "", // this property is not used
```
wkal-pubnub, line 28: what do you mean it's not used?

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/PubNubImpl.kt
```diff
@@ -179,6 +179,11 @@ open class PubNubImpl(
     val pnsdkName: String = PNSDK_PUBNUB_KOTLIN,
     eventEnginesConf: EventEnginesConf = EventEnginesConf()
 ) : PubNub {
+    internal val tokenManager: TokenManager = TokenManager()
+
+    init {
+        this.setToken(configuration.authToken)
```
wkal-pubnub, line 185: this will overwrite the `null` value in TokenManager with the default of `""` (empty string) from PNConfiguration, which will stop `authKey` (the deprecated one) from ever being used (because the order in `EndpointCore` is `tokenManager.getToken()`, then `authKey`, and `tokenManager.getToken` will now never be `null`)

I think we should make `configuration.authToken` nullable and null by default (instead of empty string) which I think will solve this. It will be different then other keys (which use empty strings), but IMHO we should over time move all keys to `String? = null` instead of `String = ""`, it's just not right to use empty strings there.

wkal-pubnub (changes requested):

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/v2/PNConfigurationImpl.kt
marcin-cebo: Ok. I will delete it.

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/v2/PNConfigurationImpl.kt
marcin-cebo: There is no way to set in Java configuration.

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/PubNubImpl.kt
marcin-cebo: Good catch. I will change the code.

wkal-pubnub (approved):

PR #312, opened 2024-11-21
Added type to setMembership

marcin-cebo:
for kotlin and java.
Added way to specify all possible includes for setMembership operation exposed by server. Changed UUID to User, uuid to userId.


pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
```diff
@@ -20,31 +20,80 @@ public static class ChannelId {
...
         return new JustChannel(new ChannelId(channelId));
     }
 
+    @Deprecated
     public static PNChannelMembership channelWithCustom(final String channelId, final Map<String, Object> custom) {
-        return new ChannelWithCustom(new ChannelId(channelId), new HashMap<>(custom));
+        return new ChannelWithCustom(new ChannelId(channelId), new HashMap<>(custom), null, null);
+    }
+
+    // PNChannelMembership.PNChannelMembership(membership01ChannelId).type("t").status("s").custom(custom).build()
```
wkal-pubnub, line 40: instantiation using a function named the same as the type is a pattern in Kotlin which has top-level functions, but not in Java. In other places in java we have a pattern like this:
`SomeType.builder()` (static function)

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
```diff
@@ -20,31 +20,80 @@ public static class ChannelId {
...
     @Getter
     private final ChannelId channel;
+    @Getter
+    private final Map<String, Object> custom;
+    @Getter
+    private final String status;
+    @Getter
+    private final String type;
 
+    @Deprecated
```
wkal-pubnub, line 30: do we need to deprecate this helper? might be still useful for someone who doesn't care about status/type

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
```diff
@@ -20,31 +20,80 @@ public static class ChannelId {
...
     public static PNChannelMembership channel(final String channelId) {
         return new JustChannel(new ChannelId(channelId));
     }
 
+    @Deprecated
```
wkal-pubnub, line 35: ditto

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
```diff
@@ -20,31 +20,80 @@ public static class ChannelId {
...
     @EqualsAndHashCode(callSuper = true)
     public static class ChannelWithCustom extends PNChannelMembership {
-        private final Object custom;
-
-        ChannelWithCustom(@NonNull final ChannelId channelId, @NonNull Object custom) {
-            super(channelId);
-            this.custom = custom;
+        ChannelWithCustom(@NonNull final ChannelId channelId,
+                          @NonNull Map<String, Object> custom,
+                          String status,
```
wkal-pubnub, line 94: this is a breaking change, leave it as it was, and in `public PNChannelMembership build()` just return an instance of PNChannelMembership:

You can try like this (I don't remember how Java works) :
`return new PNChannelMembership(channelId, custom, status, type) {};`

if that doesn't work maybe just make PNChannelMembership non-abstract or add a 
`private static class PNChannelMembershipImpl extends PNChannelMembership`

pubnub-kotlin/pubnub-kotlin-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/member/MemberInclude.kt
```diff
@@ -0,0 +1,19 @@
...
+class MemberInclude(
+    override val includeCustom: Boolean = false,
+    override val includeStatus: Boolean = false,
+    override val includeType: Boolean = false,
+    override val includeTotalCount: Boolean = false,
+    val includeUser: Boolean = false,
+    val includeUserCustom: Boolean = false,
+    val includeUserType: Boolean = false,
+    val includeUserStatus: Boolean = false,
+) : Include()
```
wkal-pubnub, line 19: do we maybe want to have some set to true by default? I don't know, but there were some in the old version that were default true

pubnub-kotlin/pubnub-kotlin-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/membership/MembershipInclude.kt
```diff
@@ -0,0 +1,99 @@
...
+        @JvmStatic
+        fun includeChannelType(includeChannelType: Boolean): Builder {
+            return Builder().includeChannelType(includeChannelType)
+        }
+
+        @JvmStatic
+        fun includeChannelStatus(includeChannelStatus: Boolean): Builder {
+            return Builder().includeChannelStatus(includeChannelStatus)
+        }
+    }
```
wkal-pubnub, line 56: not sure if this is needed, I think `fun builder(): Builder` is enough for an entry point

moreover, I don't think Kotlin users need the builder at all,
maybe experiment with `@kotlin.Deprecated(message = "JUST FOR JAVA", level = DeprecationLevel.HIDDEN)` to hide it from Kotlin but allow Java usage and see if it works?

otherwise we could have two different class hierarchies in the Kotlin and Java modules (in `com.pubnub.java.*` package)

wkal-pubnub: another thing you could try:

add a no-argument secondary constructor to `MembershipInclude` with DeprecationLevel.HIDDEN
move the methods like `fun includeCustom` onto `MembershipInclude` with DeprecationLevel.HIDDEN
remove the `Builder` class and `build()` method entirely

now from Java it should be possible to do `new MembershipInclude().includeCustom(...)` (I think?) 

the upside is that you won't have to repeat all the properties in Builder again and their default values

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/endpoints/objects/internal/IncludeQueryParam.kt
```diff
@@ -1,46 +1,55 @@
...
+        const val QUERY_PARAM_STATUS = "status"
+        const val QUERY_PARAM_CHANNEL = "channel"
+        const val QUERY_PARAM_CHANNEL_CUSTOM = "channel.custom"
+        const val QUERY_PARAM_CHANNEL_STATUS = "channel.status"
+        const val QUERY_PARAM_CHANNEL_TYPE = "channel.type"
+        const val QUERY_PARAM_UUID = "uuid"
+        const val QUERY_PARAM_UUID_CUSTOM = "uuid.custom"
+        const val QUERY_PARAM_UUID_STATUS = "uuid.status"
+        const val QUERY_PARAM_UUID_TYPE = "uuid.type"
+        const val QUERY_PARAM_INCLUDE = "include"
```
wkal-pubnub, line 31: this should have a different name than the others, as it's not one of the values, it's the key (it can be misleading now)

maybe just `QUERY_KEY_INCLUDE`

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/endpoints/objects_api/memberships/SetMembershipsImpl.java
```diff
@@ -130,4 +132,17 @@ static Collection<? extends com.pubnub.api.models.consumer.objects.PNSortKey<PNM
         }
         return list;
     }
+
+    private String getUserId() {
+        return userId != null ? userId : uuid;
+    }
+
+    private MembershipInclude getMembershipInclude() {
+        if (includeTotalCount || includeCustom || includeType) {
```
wkal-pubnub, line 141: what about includeChannel? 

maybe flip the if:
if (include != null) // new API is used
else // old API is used

pubnub-kotlin/pubnub-kotlin-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/member/MemberInclude.kt
marcin-cebo: For old I left true for includeType and includeStatus:
https://github.com/pubnub/kotlin/blob/934ec49cdea16fd524e31b93c5104c4e1ca523ae/pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/PubNubImpl.kt#L1038

https://github.com/pubnub/kotlin/blob/934ec49cdea16fd524e31b93c5104c4e1ca523ae/pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/PubNubImpl.kt#L861

For new I went with false for all. But we can change it.
What do you think?

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/endpoints/objects/internal/IncludeQueryParam.kt
marcin-cebo: Makes sense. I will change it.

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/endpoints/objects_api/memberships/SetMembershipsImpl.java
marcin-cebo: Good point. I will fix that.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
wkal-pubnub: Generally I like removing legacy code if it simplifies our codebase, but in this case I don't see much gain for us in terms of maintenance (we're not switching to some entirely different implementation, it's just a helper), but for the user it might be annoying to see deprecations and have to move to something new. 

If we can keep it because it doesn't break anything for us, then let's also keep it (undeprecated) for the user? 

Unless we really want to remove it at some point, but I don't see that we have to?

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
marcin-cebo: ok

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
marcin-cebo: Ok, I will remove deprecation

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
marcin-cebo: Ok, I will remove deprecation

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/membership/PNChannelMembership.java
marcin-cebo: ok

pubnub-kotlin/pubnub-kotlin-core-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/membership/MembershipInclude.kt
marcin-cebo: Ok, as discussed.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/endpoints/objects_api/memberships/SetMembershipsBuilder.java
```diff
@@ -0,0 +1,17 @@
+package com.pubnub.api.java.endpoints.objects_api.memberships;
+
+import com.pubnub.api.java.endpoints.Endpoint;
+import com.pubnub.api.java.endpoints.objects_api.utils.PNSortKey;
+import com.pubnub.api.java.models.consumer.objects_api.membership.MembershipInclude;
+import com.pubnub.api.java.models.consumer.objects_api.membership.PNSetMembershipResult;
+
+import java.util.Collection;
+
+public interface SetMembershipsBuilder extends Endpoint<PNSetMembershipResult> {
```
wkal-pubnub, line 10: do we need this new interface at all?
We could just make changes to `interface SetMemberships`

marcin-cebo: We don't need new interface but I thought it will be more convenient for user to deal only with valid options. There are 4 parameters in interface `SetMemberships` that should be deprecated.

What do you think?

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/pubsub/objects/PNObjectEventResult.kt
```diff
@@ -73,9 +73,10 @@ data class PNSetMembershipEvent(
     val eTag: String,
     val updated: String,
     val status: PatchValue<String?>?,
+    val type: PatchValue<String?>?,
 )
 
 data class PNDeleteMembershipEvent(
     val channelId: String,
-    val uuid: String,
+    val uuid: String, // todo do we want to change to userId?
```
marcin-cebo, line 81: Checking this.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/member/MemberInput.kt
```diff
@@ -3,7 +3,8 @@ package com.pubnub.api.models.consumer.objects.member
 import com.pubnub.kmp.CustomObject
 
 interface MemberInput {
-    val uuid: String
+    val uuid: String // todo change to userId?
```
marcin-cebo, line 6: Checking this.

pubnub-kotlin/pubnub-kotlin-api/src/commonMain/kotlin/com/pubnub/api/models/consumer/objects/member/PNMember.kt
```diff
@@ -4,17 +4,19 @@ import com.pubnub.api.models.consumer.objects.uuid.PNUUIDMetadata
 import com.pubnub.api.utils.PatchValue
 import com.pubnub.kmp.CustomObject
 
-data class PNMember(
+data class PNMember( // todo introduce new PNMember with user instead of uuid, maybe in different package
```
marcin-cebo, line 7: Checking this.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/objects_api/member/PNMembers.java
```diff
@@ -18,12 +18,13 @@
 @EqualsAndHashCode
 @ToString
 @RequiredArgsConstructor
-public class PNMembers {
+public class PNMembers { // todo dlaczego tu jest liczba mnoga a moe PNMember? Może zrobić PNMember bez uuid?
```
marcin-cebo, line 21: Checking this.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -463,44 +473,122 @@ interface PubNub : EventEmitter, StatusEmitter {
...
+        replaceWith = ReplaceWith(
+            "getMemberships(userId)"
+        )
+    )
     fun getMemberships(): GetMemberships
 
+    /**
+     * The method returns a list of channel memberships for a user. This method doesn't return a user's subscriptions.
+     */
+    fun getMemberships(userId: String): GetMembershipsBuilder
```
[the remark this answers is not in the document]
marcin-cebo: Good point. I will remove GetMembershipsBuild and reuse GetMemberships because there can't be two methods 
fun getMemberships() with different return types.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -463,44 +473,122 @@ interface PubNub : EventEmitter, StatusEmitter {
...
+        message = "Use removeMemberships(Collection<PNChannelMembership>) instead.",
+        replaceWith = ReplaceWith("removeMemberships(channelMemberships)")
+    )
     fun removeMemberships(): RemoveMemberships.Builder
 
     /**
-     * Add and remove channel memberships for a UUID.
+     * Remove channel memberships for a UUID.
+     */
+    fun removeMemberships(channelMemberships: Collection<PNChannelMembership>): RemoveMembershipsBuilder
```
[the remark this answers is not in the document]
marcin-cebo: Initially I wanted to do this to have it similar to Kotlin SDK but then I thought that maybe that this is not that bad because we can use result from getMembership and feed it to removeMemberships. Now I see that returned data from getMembership and input data to removeMemberships are not the same type (PNMembership vs. PNChannelMembership). In this situation I will change it to Collection<String>

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -463,44 +473,122 @@ interface PubNub : EventEmitter, StatusEmitter {
...
+        replaceWith = ReplaceWith("manageMemberships(channelMembershipsToAdd, channelMembershipsToRemove)")
+    )
     fun manageMemberships(): ManageMemberships.Builder
 
+    /**
+     * Add and/or remove channel memberships for a UUID.
+     */
+    fun manageMemberships(
+        channelsToSet: Collection<PNChannelMembership>,
+        channelsToDelete: Collection<PNChannelMembership>
```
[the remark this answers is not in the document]
marcin-cebo: I will change it.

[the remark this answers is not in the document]
marcin-cebo: Yes. I would leave it as is.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -463,44 +473,122 @@ interface PubNub : EventEmitter, StatusEmitter {
...
+    @Deprecated(
+        message = "Use removeChannelMembers(String, Collection<PNUUID>) instead.",
+        replaceWith = ReplaceWith("removeChannelMembers(channelId, channelMembers)")
+    )
     fun removeChannelMembers(): RemoveChannelMembers.Builder
 
+    /**
+     * Remove members from a Channel.
+     */
+    fun removeChannelMembers(channelId: String, channelMembers: Collection<PNUser>): RemoveChannelMembersBuilder
```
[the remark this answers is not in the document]
marcin-cebo: Yes, I will change it.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
```diff
@@ -463,44 +473,122 @@ interface PubNub : EventEmitter, StatusEmitter {
...
+    @Deprecated(
+        message = "Use manageChannelMembers(String, Collection<PNUUID>, Collection<PNUUID>) instead.",
+        replaceWith = ReplaceWith("manageChannelMembers(channelId, channelMembersToSet, channelMembersToRemove)")
+    )
     fun manageChannelMembers(): ManageChannelMembers.Builder
 
+    /**
+     * Set or remove members in a channel.
+     */
+    fun manageChannelMembers(channelId: String, set: Collection<PNUser>, remove: Collection<PNUser>): ManageChannelMembersBuilder
```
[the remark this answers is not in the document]
marcin-cebo: yes, I will change it.

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/endpoints/objects_api/members/SetChannelMembersImpl.java
```diff
@@ -123,4 +140,69 @@ public SetChannelMembers uuids(@NotNull final Collection<PNUUID> uuids) {
...
+        } else {
+            // if deprecated setChannelMembership API use
+            if (includeUser == Include.PNUUIDDetailsLevel.UUID) {
+                return MemberInclude.builder()
+                        .includeTotalCount(includeTotalCount)
+                        .includeCustom(includeCustom)
+                        .includeUserType(includeType)
+                        .includeStatus(true)
+                        .includeUser(true)
+                        .build();
```
wkal-pubnub, line 187: instead of repeating the builder 3 times, it'll be more concise to have a variable to hold the builder and then

```
if (includeUser == Include.PNUUIDDetailsLevel.UUID) {
    builder.includeUser(true)
}
if (includeUser == Include.PNUUIDDetailsLevel.UUID_WITH_CUSTOM) {
    builder.includeUser(true)
    builder.includeUserCustom(true)
}
return builder.build()
```

marcin-cebo: Good point. I will refactor it.

wkal-pubnub (approved):

pubnub-gson/pubnub-gson-impl/src/main/java/com/pubnub/internal/java/endpoints/objects_api/members/GetChannelMembersImpl.java
```diff
@@ -19,19 +21,22 @@
...
     private final String channel;
     private Integer limit = null;
     private PNPage page;
     private String filter;
     private Collection<PNSortKey> sort = Collections.emptyList();
-    private boolean includeTotalCount;
-    private boolean includeCustom;
-    private boolean includeType;
-    private Include.PNUUIDDetailsLevel includeUUID;
+    private boolean includeTotalCount; // deprecated
```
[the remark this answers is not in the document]
marcin-cebo: This is the case:
```
pubNubUnderTest.setChannelMembers()
                .channel(testChannelId)
                .uuids(channelMembers)
                .includeTotalCount(true)
                .includeCustom(true)
                .includeUUID(UUID_WITH_CUSTOM)
                .sync();
```
There is deprecation on setChannelMembers() so I thought this would be enough. I didn't want to pollute code by adding e.g.:

```
    /**
     * @deprecated Use {@link #include} instead.
     */
    @Deprecated
    private boolean includeTotalCount;
```
I can add this. What do you think?

PR #314, opened 2024-11-26
Isolate exceptions in event/status listeners

wkal-pubnub:
An exception in a user's listener shouldn't crash the whole event engine


pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/managers/ListenerManager.kt
```diff
@@ -65,49 +65,59 @@ class ListenerManager(val pubnub: PubNub) : MessagesConsumer, StatusConsumer, Ev
...
+        subscriptionCallbacks.safeForEach { it.file(pubnub, envelope) }
+        setCallbacks.safeForEach { it.file(pubnub, envelope) }
+    }
+
+    private inline fun <T> Iterable<T>.safeForEach(action: (T) -> Unit) {
+        for (element in this) {
+            try {
+                action(element)
+            } catch (e: Throwable) {
+                e.printStackTrace()
```
marcin-cebo, line 118: What about using logging instead of printStackTrace()?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/managers/ListenerManager.kt
wkal-pubnub: done

PR #315, opened 2024-11-26
Commonize CustomObject handling for JS/iOS

wkal-pubnub:


pubnub-kotlin/pubnub-kotlin-core-api/src/jvmMain/kotlin/com/pubnub/api/v2/PNConfiguration.jvm.kt
```diff
@@ -45,6 +45,6 @@ actual fun createPNConfiguration(
         this.authKey = NO_AUTH_KEY
         this.secretKey = secretKey.orEmpty()
         this.logVerbosity = logVerbosity
-        this.authToken = authToken.orEmpty()
+        this.authToken = authToken
```
marcin-cebo, line 48: What is a reason for this change?

marcin-cebo (approved):

pubnub-kotlin/pubnub-kotlin-core-api/src/jvmMain/kotlin/com/pubnub/api/v2/PNConfiguration.jvm.kt
wkal-pubnub: we changed in to nullable at some point so it shouldn't be assigned an empty string

PR #323, opened 2024-12-13
Updated libs and libs info.

marcin-cebo:


.pubnub.yml
```diff
@@ -66,45 +66,52 @@ sdks:
...
                       - Windows Vista Ultimate
                     maximum-os-version:
-                      - Windows 10 Home
+                      - Windows 11 Home
                     target-architecture:
                       - x86
                       - x86-64
               requires:
+                -
+                  name: kotlin
```
wkal-pubnub, line 75:
```suggestion
                  name: kotlin-stdlib
```

.pubnub.yml
```diff
@@ -66,45 +66,52 @@ sdks:
...
+                  min-version: 2.0.21
```
wkal-pubnub, line 76: we use 2.0.21, but we're compatible with at least 1.8.0 and above, which I think was the point to capture this information here

wkal-pubnub (approved):

PR #328, opened 2025-01-17
Etag support for users and channels

wkal-pubnub:
feat: A new optional parameter `ifMatchesEtag` is added to `setUUIDMetadata` and `setChannelMetadata`. When provided, the server compares the argument value with the ETag on the server and if they don't match a HTTP 412 error is returned.


pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/endpoints/objects_api/channel/SetChannelMetadata.java
```diff
@@ -19,6 +20,14 @@ public interface SetChannelMetadata extends Endpoint<PNSetChannelMetadataResult>
 
     SetChannelMetadata includeCustom(boolean includeCustom);
 
+    /**
+     * Optional entity tag from a previously received `PNChannelMetadata`. The request
+     * will fail if this parameter is specified and the ETag value on the server doesn't match.
+     * @param etag from PNChannelMetadata
+     * @return this builder
+     */
+    SetChannelMetadata ifMatchesEtag(@Nullable String etag);
```
marcin-cebo, line 29: Wouldn't be enough to name it just `eTag` ?  What is a value of having `ifMatches` in the name?

marcin-cebo (approved):

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/endpoints/objects_api/channel/SetChannelMetadata.java
wkal-pubnub: if we name it `eTag` then it's exactly like the other fields that we are updating: the value in `name` updates `name` in the User on the server, `description` updates `description`, and so `eTag` would update the `eTag`? ... but we can't update `eTag` because it's not updatable, it's generated by the server. 
So I wanted to make this clear that that it's a condition - make this request if the etag matches the passed in value. - ifMatchesEtag for short

marcin-cebo: ok, makes sense.

wkal-pubnub:
@pubnub-release-bot release kotlin as v10.4.0

PR #331, opened 2025-01-24
pubNub.forceDestroy should force okHttpClient executor service to shoutDown immediately.

marcin-cebo:
fix: pubNub.forceDestroy should cause okHttpClient executor service to shutDown immediately.


jguz-pubnub (approved):

marcin-cebo:
@pubnub-release-bot release kotlin as v10.4.1

PR #339, opened 2025-03-10
Mc/matchmaking poc1

marcin-cebo:


pubnub-matchmaking-kotlin/pubnub-matchmaking-kotlin-impl/src/commonMain/kotlin/com/pubnub/matchmaking/internal/serverREST/MatchmakingRestService.kt
```diff
@@ -0,0 +1,273 @@
...
+            0.0
+        } else {
+            regionalPriority
+        }
+
+        return skillGapWeight * eloDifference + regionMismatchPenalty
+    }
+
+    // Create all possible pairs (only including allowed pairs)
+    private fun createAllPairs(users: List<User>): List<Triple<Int, Int, Double>> {
```
wkal-pubnub, line 212: this List<Triple<Int,Int,Double>> is unreadable (it's difficult to say what these ints and doubles represent), consider a better return type

pubnub-matchmaking-kotlin/pubnub-matchmaking-kotlin-impl/src/commonMain/kotlin/com/pubnub/matchmaking/internal/serverREST/MatchmakingRestService.kt
```diff
@@ -0,0 +1,273 @@
...
+        val regionalPriority = constraints["REGIONAL_PRIORITY"] as Double
+
+        val eloA = (userA.custom?.get("elo") as? Int) ?: 0
+        val eloB = (userB.custom?.get("elo") as? Int) ?: 0
+        val regionA = (userA.custom?.get("server") as? String) ?: "global"
+        val regionB = (userB.custom?.get("server") as? String) ?: "global"
+
+        val eloDifference = abs(eloA - eloB)
+        if (eloDifference > maxEloGap) {
+            return Double.POSITIVE_INFINITY
```
wkal-pubnub, line 199: wouldn't `null` instead of `Double.POSITIVE_INFINITY` be more appropriate in this case?

pubnub-matchmaking-kotlin/pubnub-matchmaking-kotlin-impl/src/commonMain/kotlin/com/pubnub/matchmaking/internal/serverREST/MatchmakingRestService.kt
```diff
@@ -0,0 +1,273 @@
...
+                    pairs.add(Triple(i, j, score))
+                }
+            }
+        }
+        return pairs
+    }
+
+    // Greedy pairing: sort by score and select pairs without conflicts.
+// Instead of returning MatchmakingPair, we create a MatchGroup that holds a set of users.
+    private fun pairUsersBySkill(users: List<User>): MatchmakingResult {
```
wkal-pubnub, line 227: in general I think this algorithm is wasteful: 
1) why create all pairs when we're only going to use some of them?
2) and why do it every 5 seconds, even if nothing has changed (no new users are requesting a match)?

instead of running a loop that checks all possibilities every X seconds, I'd explore something more event-based, like:

1) there is some `List<MatchGroup>` that is initially empty
2) a user with some `userId` and `skill` calls `findMatch()`
2) `findMatch` iterates over the list of open match groups (i.e. matches that have not started because they don't have the required number of players) and if the `skill` matches the group's `skill` level (due to some rules), add the user to the group
3) if the group has the necessary number of players, inform all users in the group that they have a match (using the callback, or in REST by returning HTTP 200)
4) if there is no group that matches the player's skill, open a new group with this user and add it to the list of open matches (HTTP is in progress until the group is complete)

So everything would happen in response to a new player requesting a match. The only timer that I'd add would be to support some sort of timeout if the player has been waiting for too long to return with a "no match found" error or something like that.

pubnub-matchmaking-kotlin/pubnub-matchmaking-kotlin-impl/src/commonMain/kotlin/com/pubnub/matchmaking/internal/serverREST/MatchmakingRestServiceNew.kt
```diff
@@ -0,0 +1,157 @@
...
+        var groupToJoin: OpenMatchGroup?
+
+        groupsMutex.withLock {
+            // Find an open group where the user's skill is compatible.
+            groupToJoin = openMatchGroups.firstOrNull { group ->
+                isSkillCompatible(user, group) && group.users.size < group.requiredSize
+            }
+            if (groupToJoin != null) {
+                // Join the found group.
+                groupToJoin!!.users.add(user)
```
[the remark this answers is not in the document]
wkal-pubnub: or just make it a `val` instead of `var` so you can use smart cast

PR #348, opened 2025-05-28
added basic usage

techwritermat:


pubnub-kotlin/pubnub-kotlin-docs/src/main/kotlin/com/pubnub/docs/publishAndSubscribe/subscribe/SubscribeOld.kt
```diff
@@ -28,6 +28,19 @@ class SubscribeOld : SnippetBase() {
         // snippet.end
     }
 
+    private fun subscribe() {
+        // https://www.pubnub.com/docs/sdks/kotlin/api-reference/publish-and-subscribe#basic-usage-1
```
marcin-cebo, line 32: This url:
https://www.pubnub.com/docs/sdks/kotlin/api-reference/publish-and-subscribe#basic-usage-1
refers to 

```
val config = com.pubnub.api.v2.PNConfiguration.builder(UserId("myUserId"), "demo").apply {
    publishKey = "demo"
}

val channel = pubnub.channel("myChannel")

channel.fire(myMessage).async { result ->
    result.onFailure { exception ->
        println("Error while publishing")
        exception.printStackTrace()
    }.onSuccess { value ->
        println("Message sent, timetoken: ${value.timetoken}")
    }
}
```
that has been placed in FireOthers#fireBasic

PR #373, opened 2025-10-02
Added log message about deprecated method usage.

marcin-cebo:


pubnub-kotlin/pubnub-kotlin-docs/src/main/kotlin/com/pubnub/docs/logging/customLogger/MonitoringService.kt
```diff
@@ -0,0 +1,22 @@
...
+class MonitoringService {
+    fun reportError(
+        message: Any,
+        stackTrace: List<String>?,
+        errorType: String?,
+        instanceId: String?,
+        timestamp: String,
+        location: String?
+    ) {
+        // no need to implement
```
jguz-pubnub, line 12: The same here, do we need thee `// no need to implement` comments?

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/logging/CompositeLogger.kt
```diff
@@ -38,6 +39,11 @@ class CompositeLogger(
     }
 
     override fun debug(message: LogMessage) {
+        // todo remove
```
jguz-pubnub, line 42: Should it be removed?

pubnub-gson/pubnub-gson-docs/src/main/java/com/pubnub/docs/logging/customLogger/AnalyticsService.java
```diff
@@ -0,0 +1,10 @@
+package com.pubnub.docs.logging.customLogger;
+
+import java.util.Map;
+
+public class AnalyticsService {
+
+    public void trackConfigChange(String instanceId, Map<String, Object> configDetail) {
+        //no need to implement
```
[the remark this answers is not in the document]
marcin-cebo: This is part of codeSnippet i.e. AnalyticsService is used in MonitoringLogger that will be presented in logs.
I added this comment to indicate that this is conscious decision to left implementation empty

pubnub-kotlin/pubnub-kotlin-impl/src/main/kotlin/com/pubnub/internal/logging/CompositeLogger.kt
marcin-cebo: Good catch. I will remove it.

PR #412, opened 2026-07-21
Extend grantToken and parseToken API to be used in DataSync

marcin-cebo:


pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
```diff
@@ -83,4 +89,171 @@ public void happyPath() throws PubNubException {
                 pnToken.getPatterns().getChannelGroups().get(expectedChannelGroupPattern));
     }
 
+    @Test
+    public void happyPath_datasync() throws PubNubException {
```
jguz-pubnub: Did you consider names other than `happyPath_datasync`? It's very vague

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
```diff
@@ -83,4 +89,171 @@ public void happyPath() throws PubNubException {
...
+        // composite keys the SDK is expected to build: "<namespace>:<id>"
+        final String entityKey = DataSyncGrant.DATASYNC_ENTITIES + ":" + entityId;
+        final String entityPatternKey = DataSyncGrant.DATASYNC_ENTITIES + ":" + entityPatternId;
+        final String relationshipKey = DataSyncGrant.DATASYNC_RELATIONSHIPS + ":" + relationshipId;
+        final String relationshipPatternKey = DataSyncGrant.DATASYNC_RELATIONSHIPS + ":" + relationshipPatternId;
+        final String membershipKey = DataSyncGrant.DATASYNC_MEMBERSHIPS + ":" + membershipId;
+        final String membershipPatternKey = DataSyncGrant.DATASYNC_MEMBERSHIPS + ":" + membershipPatternId;
+
+        // when
+        final PNGrantTokenResult grantTokenResponse = pubNubUnderTest
```
jguz-pubnub, line 158: I wonder what the whole test will look like if you put inline strings rather than extracting them into local variables. Have you tried it in action? Would it be more readable and much shorter? For example:

```
DataSyncGrant.entity( "user.A").get().update().projection("admin")
DataSyncGrant.entityPattern("user.*").get().projection("__default__"),
// etc.
```

I leave it up to you; what I wanted to signalize is whether inline strings make sense in your opinion. Here and in all other tests

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
```diff
@@ -83,4 +89,171 @@ public void happyPath() throws PubNubException {
...
+        PubNub pubNubUnderTest = getServer();
+        final int expectedTTL = 1337;
+        final String entityName = "capy-001";
+        final String membershipName = "user-123:channel-X";
+
+        //when — full fluent Java-facing DataSync grant chain
+        final PNGrantTokenResult grantTokenResponse = pubNubUnderTest
+                .grantToken(expectedTTL)
+                .authorizedUUID("pam-debug-admin")
+                .datasync(Arrays.asList(
```
jguz-pubnub: Should the builder method be named `dataSync` rather than `datasync`?

jguz-pubnub:
There's a `datasync` literal in many places. The question is: should it rather be camel-cased `dataSync` instead?

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
marcin-cebo: I would rename it to:
can_grantToken_for_datasync_resources_and_patterns

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
marcin-cebo: Using strings rather than extracting them into local variable would make test shorter but the thing is that the variables are needed later in a test for assertions e.g.

assertEquals(adminProjection, typedRes.getEntities().get(entityId));
assertEquals(adminProjection, typedRes.getRelationships().get(relationshipId));

What do you think?

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/pam/GrantTokenIT.java
marcin-cebo: Good point. I will rename it.

PR #413, opened 2026-07-28
Add DataSync Entities APIs

marcin-cebo:
Added create/get/remove entity APIs for Kotlin/Java
Added update/getAll/patch entity APIs for Kotlin/Java

### DIFF
--- a/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
+++ b/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
@@ -486,7 +486,7 @@ interface PubNub : EventEmitter, StatusEmitter {
     fun removeChannelMetadata(): RemoveChannelMetadata.Builder
 
     /**
-     * Entry point for the DataSync (App Context v4) API.
+     * Entry point for the DataSync API.
      */
     fun dataSync(): DataSync
 
--- a/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
+++ b/pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
@@ -7,6 +7,7 @@ import com.pubnub.api.java.builder.PresenceBuilder
 import com.pubnub.api.java.builder.SubscribeBuilder
 import com.pubnub.api.java.builder.UnsubscribeBuilder
 import com.pubnub.api.java.callbacks.SubscribeCallback
+import com.pubnub.api.java.datasync.DataSync
 import com.pubnub.api.java.endpoints.DeleteMessages
 import com.pubnub.api.java.endpoints.FetchMessages
 import com.pubnub.api.java.endpoints.History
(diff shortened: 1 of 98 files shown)


pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/datasync/entity/PNEntity.java
```diff
@@ -0,0 +1,44 @@
...
+@Accessors(chain = true)
+@ToString
+public class PNEntity {
+    private String id;
+    private String entityClass;
+    private int entityClassVersion;
+
+    /**
+     * The level at which the entity class is defined (e.g. {@code SubKey} / {@code Global}).
+     * Modeled as a nullable {@link String} rather than an enum to stay source-compatible with future values.
```
jguz-pubnub: Should a comment like this be present?

```
//  * Modeled as a nullable {@link String} rather than an enum to stay source-compatible with future values.
//

File: pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt [hunk 2]
```diff
@@ -484,6 +485,11 @@ interface PubNub : EventEmitter, StatusEmitter {
      */
     fun removeChannelMetadata(): RemoveChannelMetadata.Builder
 
+    /**
+     * Entry point for the DataSync (App Context v4) API.
```
jguz-pubnub: This isn't my decision. But... should we ever mix AppContext with DataSync?

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/datasync/EntityIT.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class EntityIT extends BaseIntegrationTest {
```
jguz-pubnub: What's wrong with having `EntityIntegrationTest` as a class name instead of `EntityIT`? Did you encounter any linter (static code analysis) errors?

jguz-pubnub (approved):

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/models/consumer/datasync/entity/PNEntity.java
marcin-cebo: I will remove it.

pubnub-gson/pubnub-gson-api/src/main/java/com/pubnub/api/java/PubNub.kt
marcin-cebo: Good point. I will remove "(App Context v4)" here and in a few other places.

pubnub-gson/pubnub-gson-impl/src/integrationTest/java/com/pubnub/api/integration/datasync/EntityIT.java
marcin-cebo: Some integration tests in Java uses ..IT.java convention but it seems most of them has ....IntegrationTest.java suffix. I will rename this test to DataSyncEntityIntegrationTest.java to match equivalent test in Kotlin

